PR #7, opened 2023-07-28
Add maven p2 repository

pstuecker:
Adds an experimental maven p2 repository which may help us to deploy a p2 repo to github packages.


TruongQuangSB (approved):
ok

PR #11, opened 2023-08-09
Refactoring SWTBot

TruongQuangSB:
- Use ParameterizedTest to avoid same Test class


java/bundles/org.eclipse.set.swtbot/src/org/eclipse/set/swtbot/table/TableDataTest.java
```diff
@@ -0,0 +1,80 @@
...
+	/**
+	 * Compare table data with reference file
+	 * 
+	 * @throws Exception
+	 */
+	@ParameterizedTest
+	@MethodSource("providesPtTable")
+	void testTableData(final PtTable table) throws Exception {
+		givenNattableBot(table);
+		whenNattableBotNotNull();
```
pstuecker, line 75:
```suggestion
```
We do not need to test the test here.

pstuecker (changes requested):

java/bundles/org.eclipse.set.swtbot/src/org/eclipse/set/swtbot/table/TableDataTest.java
TruongQuangSB: ok

pstuecker (approved):

PR #4, opened 2023-07-13
Implementation Layoutinformation loading, when exists

TruongQuangSB:
- This PR should merge after GithubAction for SET implemented
- Load PlanPro file with layout.xml

PR #15, opened 2023-08-14
Validate Layoutinformations

TruongQuangSB:
- The [PR-set#4](https://github.com/eclipse-set/set/pull/4) should be merge before this PR
- Validate Layoutinformationen and show problem in validation view


java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/dialogservice/ReportSavedSessionDialog.java
```diff
@@ -76,21 +76,18 @@ public ReportSavedSessionDialog(final Shell parent,
 
 	private static String getMessage(final IModelSession session,
 			final Messages messages) {
-		final Outcome outcome = session.getValidationResult().getOutcome();
+		final Outcome outcome = session
+				.getValidationsOutcome(ValidationResult::getOutcome);
 		final Path location = session.getToolboxFile().getPath();
 		switch (outcome) {
-		case VALID:
+		case VALID, NOT_SUPPORTED:
```
pstuecker, line 83: This behavior shouldn't be changed here.
```suggestion
		case VALID:
```

TruongQuangSB: you are right

pstuecker (approved):

PR #34, opened 2023-08-31
Load Layoutinformation separate

TruongQuangSB:
- Layoutinformation load now separate with PlanPro Model
- Fix validate problem show false object state


PetersNils:
@TruongQuangSB the PlainToolboxFileTest is failing, this should be fixed before merging.

TruongQuangSB:
> @TruongQuangSB the PlainToolboxFileTest is failing, this should be fixed before merging.

fixed

PR #80, opened 2023-09-18
Fix jump function, when initial state empty

TruongQuangSB:
- Fix problem: when table in diff state and current modell include empty initial state, then jump function from table to object in text view doesn't work.

### DIFF
--- a/web/textviewer/src/app.ts
+++ b/web/textviewer/src/app.ts
@@ -8,7 +8,7 @@
  */
 import { Model, ProblemMessage, TextFileModel } from './model'
 import * as monaco from 'monaco-editor'
-import jumpToDefinition, { ModelContainer, findObjectDefinitionLineByGUID, getGUIDAt, getModelContainer } from './jumpToGuid'
+import jumpToDefinition, { ModelContainer, findObjectDefinitionLineByGUID, getGUIDAt, getModelContainer, switchModelToFindReference } from './jumpToGuid'
 import { TableType } from '.'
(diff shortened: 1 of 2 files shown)


web/textviewer/src/app.ts
```diff
@@ -136,15 +136,31 @@ export class App {
...
     if (line) {
       // Jump to the line
       this.editor.revealLineInCenter(line)
       this.editor.setPosition({
         lineNumber: line,
         column: 0
       })
+      return
     }
+    switchModelToFindReference(this.editor, this.viewModels, guid)
```
TruongQuangSB, line 163: change to layout model when guid in technical model aren't found

web/textviewer/src/app.ts
```diff
@@ -136,15 +136,31 @@ export class App {
...
+    let line = findObjectDefinitionLineByGUID(guid, this.editor.getModel(), this.xml, modelContainer)
+    if (!line && sessionState === TableType.DIFF) {
+      switch (modelContainer) {
+        case ModelContainer.INITIAL:{
+          line = findObjectDefinitionLineByGUID(guid, this.editor.getModel(), this.xml, ModelContainer.FINAL)
+          break
+        }
+        case ModelContainer.FINAL:{
+          line = findObjectDefinitionLineByGUID(guid, this.editor.getModel(), this.xml, ModelContainer.INITIAL)
+          break
```
TruongQuangSB, line 149: change to another model state, when table is diff state and can't found guid in current state, where editor cursor stay

PR #83, opened 2023-09-20
Fix SWTBot TableDataTest

TruongQuangSB:
- TableDataTest was only check the data on current viewport, not total the table.


PetersNils (changes requested):
As far as I understand your changes this check could still miss an mismatch to the reference file. Consider a case in which the reference CSV contains more rows or colums than colums/rows that are present in the application. As the current test implementation is iterating on the Nattable excess rows/colums in the reference file are never tested for.

I think a good way to check for this is in an extra test that enforces an exact match on both column and row sizes.

TruongQuangSB:
> As far as I understand your changes this check could still miss an mismatch to the reference file. Consider a case in which the reference CSV contains more rows or colums than colums/rows that are present in the application. As the current test implementation is iterating on the Nattable excess rows/colums in the reference file are never tested for.
> 
> I think a good way to check for this is in an extra test that enforces an exact match on both column and row sizes.
you are right

PetersNils (approved):

PR #90, opened 2023-09-27
Grouping same content row in Validation- and PlazModel view

TruongQuangSB:
- Grouping same problem (index, linenumber, description will not considered) in validation- and plazmodel view
- By default will all grouping collapsed
- Configuration minimun element of a group in config.ini with parameter : toolbox.table.tree.minimum


java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/sorting/AbstractSortByColumnTables.java
```diff
@@ -91,13 +105,20 @@ protected NatTable createTable(final Composite parent,
...
 				bodyLayerStack.getSelectionLayer());
+
+		// Sort Column Header
+		final TableSortModel sortModel = new TableSortModel(bodyDataProvider);
 		final SortHeaderLayer<BodyLayerStack> sortHeaderLayer = new SortHeaderLayer<>(
-				columnHeaderLayer, new TableSortModel(bodyDataProvider), true);
+				columnHeaderLayer, sortModel, true);
+		sortHeaderLayer.unregisterCommandHandler(SortColumnCommand.class);
+		sortHeaderLayer.registerCommandHandler(new TreeSortCommandHandler(
+				sortModel, sortHeaderLayer, treeLayer));
```
TruongQuangSB, line 121: unregister default commandhandler and add custom handler

pom.xml
```diff
@@ -166,7 +166,7 @@
 			<plugin>
 				<groupId>org.eclipse.xtend</groupId>
 				<artifactId>xtend-maven-plugin</artifactId>
-				<version>2.27.0</version>
+				<version>2.32.0</version>
```
pstuecker, line 169:
```suggestion
				<version>2.27.0</version>
```
Xtend requires Xtend libs, which aren't updated yet (as they depend on a new Eclipse platform).

pstuecker (changes requested):

pom.xml
TruongQuangSB: ok, i have upgrade,  because 2.27 version not working with default function in interface. Now I have IMPROVE for this default function notice

pstuecker (approved):

PR #95, opened 2023-09-27
Setup CodeQL

pstuecker:
For evaluation, only scan main branch.

Also migrate to common Java setup and avoid rebuilds on DEPENDENCIES updates

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,11 @@
 
 name: Build SET
 
-on: [push, repository_dispatch]
+on: 
+  push:
+    paths-ignore:
+      - DEPENDENCIES
+  repository_dispatch:
 
 env:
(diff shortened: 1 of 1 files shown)


TruongQuangSB (approved):

.github/workflows/ci.yml
```diff
@@ -2,7 +2,11 @@
 
 name: Build SET
 
-on: [push, repository_dispatch]
+on: 
+  push:
+    paths-ignore:
+      - DEPENDENCIES
+  repository_dispatch:
```
TruongQuangSB, line 9: i think another repo should be ignore DEPENDENCIES too

pstuecker: Yes. I'll do that when we add CodeQL or something else next.

PR #100, opened 2023-09-28
Add siteplan implementation

pstuecker:
This adds the siteplan component used for visualizing the planning data contained in a .planpro file. It also integrates a small developer help which helps in configuring various parameters (e.g. default zoom level) for the siteplan. 

## Note

As this is a large change with more than 1000 lines, we require an additional IP Review per the Eclipse Development Process.


java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/ToolboxProperties.java
```diff
@@ -57,7 +57,7 @@ public interface ToolboxProperties {
 	public static final String GREYED_OUT_STYLE_FOREGROUND = "toolbox.greyedoutstyle.foreground";
 
 	/**
-	 * Property key for log file (empty name deactivates logging)
+	 * Property key for log file = empty name deactivates logging)
```
pstuecker, line 60:
```suggestion
	 * Property key for log file (empty name deactivates logging)
```

pstuecker:
#101

PR #119, opened 2023-10-16
Fix incorrect validation results when missing layout xml

pstuecker:
If a planning data has no layout information, the XSD validation was always considered to have failed due to the default value in the model. Furthermore it was not possible to differenciate between success messages of the model XSD validation and the layout XSD validation.

This PR disables validation for missing layout information xml files and adds the Layout state tag to the success entries.


java/bundles/org.eclipse.set.feature.validation/src/org/eclipse/set/feature/validation/session/ModelSession.java
```diff
@@ -15,7 +15,6 @@
 import java.util.Collections;
 import java.util.EventObject;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
```
TruongQuangSB, line 18:
```suggestion
import java.util.Map;
import java.util.Objects;
```

TruongQuangSB (approved):

PR #123, opened 2023-10-20
Fix getFstrZugCrossingLeg considering Fstrs outside the associated Bereich Objekt

pstuecker:
When determining the Fstr_Fahrweg for a given track switch crossing leg, we only considered the TOP_Kante to check if the track switch is included within the Fstr_Fahrweg. However this also incorrectly considered the case where a Fstr_Fahrweg starts at some point after the track switch leg.

### DIFF
--- a/java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/WKrGspElementExtensions.xtend
+++ b/java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/WKrGspElementExtensions.xtend
@@ -185,7 +185,7 @@ class WKrGspElementExtensions extends BasisObjektExtensions {
 			fstrZug !== null && IDFstrFahrweg.bereichObjektTeilbereich.map [
 				topKante
 			].contains(legTopKante) && element.getWKrGspKomponenten.exists [
-				fstrZR.IDFstrFahrweg.intersects(it)
+				fstrFW.intersects(it)
 			]
 		].filterNull.toSet
 	}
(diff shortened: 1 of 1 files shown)


java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/WKrGspElementExtensions.xtend
```diff
@@ -182,9 +181,12 @@ class WKrGspElementExtensions extends BasisObjektExtensions {
 	static def Set<Fstr_Zug_Rangier> getFstrZugCrossingLeg(
 		W_Kr_Gsp_Element element, TOP_Kante legTopKante) {
 		return element.container.fstrZugRangier.filter [
+			val fstrZR = it
```
TruongQuangSB, line 184:
```suggestion
			val fstrFW = IDFstrFahrweg
```

java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/WKrGspElementExtensions.xtend
```diff
@@ -182,9 +181,12 @@ class WKrGspElementExtensions extends BasisObjektExtensions {
...
 			fstrZug !== null && IDFstrFahrweg.bereichObjektTeilbereich.map [
 				topKante
-			].contains(legTopKante)
+			].contains(legTopKante) && element.getWKrGspKomponenten.exists [
+				fstrZR.IDFstrFahrweg.intersects(it)
```
TruongQuangSB, line 188:
```suggestion
				fstrFW.intersects(it)
```

PR #132, opened 2023-10-26
Fix incorrect initial/final state comparision for Sslw

pstuecker:


pstuecker:
See https://github.com/eclipse-set/set/pull/136

PR #223, opened 2023-11-22
Fix Exception By Project data editor

TruongQuangSB:
Problem: 
- It was set all resources (technial and layout) with same URI.
- Illegal reference object was set by fill default.

### DIFF
--- a/java/bundles/org.eclipse.set.feature.validation/src/org/eclipse/set/feature/validation/session/ModelSession.java
+++ b/java/bundles/org.eclipse.set.feature.validation/src/org/eclipse/set/feature/validation/session/ModelSession.java
@@ -751,24 +751,19 @@ public class ModelSession implements IModelSession {
 			}
 			PlanProSchnittstelleExtensions.fix(getPlanProSchnittstelle());
 			saveFixResult = SaveFixResult.NONE;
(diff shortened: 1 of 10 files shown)


java/bundles/org.eclipse.set.feature.validation/src/org/eclipse/set/feature/validation/session/ModelSession.java
```diff
@@ -751,24 +751,19 @@ private boolean savePlanPro(final Shell shell, final boolean askUser)
...
 					getToolboxFile().getEditingDomain());
-			// Ask user if all missing values should be filled
-			if (PlanProSchnittstelleExtensions.containsUnfilledValues(
-					getPlanProSchnittstelle(), toolboxFile.getPlanProResource())
-					&& serviceProvider.dialogService
-							.confirmSetDefaultsGlobally(shell)) {
-				PlanProSchnittstelleExtensions
-						.fixDefaults(getPlanProSchnittstelle());
-
-				saveFixResult = SaveFixResult.GLOBAL;
```
TruongQuangSB, line 770: to avoid set reference to illegal object, we should only automatic fill by ManagementObject

pstuecker (approved):

PR #249, opened 2023-12-08
Fix parser trackswitch form from csv, when the form contain ","

TruongQuangSB:


java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/TrackSwitchMetadataProvider.java
```diff
@@ -74,27 +74,37 @@ public TrackSwitchMetadata getTrackSwitchMetadata(final ENUMWKrArt type,
...
+		return result;
+	}
+
+	private static TrackSwitchMetadata getTrackSwitchMetadata(
+			final List<TrackSwitchMetadata> md, final String speziellDesign) {
 		for (final TrackSwitchMetadata entry : md) {
-			if (designWithPrefix
-					.equals(entry.getDesignString().replace(" ", ""))) { //$NON-NLS-1$ //$NON-NLS-2$
+			if (speziellDesign
+					.equals(entry.getDesignString().replaceAll(" |\"", ""))) { //$NON-NLS-1$ //$NON-NLS-2$
```
pstuecker, line 104: The replaceAll here is surprising. I'd move it to the calling function.

pstuecker (changes requested):

java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/TrackSwitchMetadataProvider.java
TruongQuangSB: The replaceAll here replace by csv file and by calling function replace by model

PR #252, opened 2023-12-11
Fix siteplan exception if label is nil

pstuecker:


TruongQuangSB (approved):

TruongQuangSB (changes requested):
see comments

web/siteplan/src/model/SvgElement.ts
```diff
@@ -91,7 +91,7 @@ export class SvgBridgeSignal extends SvgElement {
     nullpunkt: ISvgPoint | null,
     mountOffset: number,
     mountDirection: MountDirection,
-    signalLabel: Label | null,
+    signalLabel: Label | undefined,
```
TruongQuangSB, line 94:
```suggestion
    signalLabel?: Label,
```
for consistent with another label attribute

TruongQuangSB (approved):

PR #320, opened 2024-01-15
Improve topological graph finding

pstuecker:
This PR introduces a new TopologicalGraphService based on JGraphT for finding  distances between points on the topological track graph.

Compared to our existing Digraphs/TopRouting implementation this grants us the following benefits:

- Performance is significantly better (~20% speed up for PPHN Sskp generation)
- The TOPGraph is kept during a session rather than recreating the graph multiple times
- We can search for arbitrary points on a TopEdge (TopRouting only allows to search for point objects). 

Limitations:

- Currently there is no support for the point object path iterator, so we can't remove the Digraph implementation yet. 

The new graph service is now being used for most relevant tables as well as for a new implementation for the Sskp Bahnsteig<-> PZB distances.

This also fixes an old issue with the CacheService not being usable from OSGi services.


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskp/SskpBahnsteigUtils.java
```diff
@@ -0,0 +1,189 @@
...
+		public OptionalDouble getDistanceEnd() {
+			return distanceEnd;
+		}
+	}
+
+	private static boolean isInBOTB(
+			final Bereich_Objekt_Teilbereich_AttributeGroup botb,
+			final Punkt_Objekt_TOP_Kante_AttributeGroup point) {
+		final double limitA = botb.getBegrenzungA().getWert().doubleValue();
+		final double limitB = botb.getBegrenzungA().getWert().doubleValue();
```
TruongQuangSB, line 64:
```suggestion
		final double limitB = botb.getBegrenzungB().getWert().doubleValue();
```

TruongQuangSB (changes requested):

pstuecker:
Changes are applied in https://github.com/eclipse-set/set/pull/365

TruongQuangSB (approved):

PR #235, opened 2023-11-29
Overviewplan: Track Transform

TruongQuangSB:
- The [PR#206](https://github.com/eclipse-set/set/pull/206) should be merge first.
- Add Track Transformator:
  - Currently the transformator only find out the overview track, which contains a lot TOPKanten that represent a straight line in the Overviewplan

PR #206, opened 2023-11-17
Implement Overviewplan infrastructure

TruongQuangSB:
- Adding Overviewplan bundles.
- Implement Overviewplan part (exist only in developmode).

PR #354, opened 2024-01-31
Visualation Overviewplan Track

TruongQuangSB:
- #206 and #235 should merge first
- In this pull request:
  - Find TOP_Kante default length through all connect path of this TOP_Kante.
  - Find Track level through combine left/right tracks.
  - Avoid Track/TOP_Kante overlap through increase/decrease level of the track.
  - Visualation Overviewplan track.
  - Each TOP_Kante have color same Siteplan for easy to detect, where the TOP_Kante in siteplan lie.


java/bundles/org.eclipse.set.feature.overviewplan/src/org/eclipse/set/feature/overviewplan/parts/WebOverviewplanPart.java
```diff
@@ -36,10 +39,13 @@
  */
 public class WebOverviewplanPart extends BasePart {
 
-	private static final String WEB_PATH = "./web/siteplan"; //$NON-NLS-1$
+	private static final String WEB_PATH = "./web/overviewplan"; //$NON-NLS-1$
```
pstuecker, line 42: The frontend code should be shared between overview plan and siteplan

pstuecker (changes requested):

pstuecker (approved):

PR #365, opened 2024-02-05
Update banking value calculations

pstuecker:
This updates the method used for calculating bank values for signals.

Old method:

- For each signal, search to either side to find the nearest `Ueberhoehung`. Then attempt to match the two found points to a `Ueberhoehungslinie` or default to a linear calculation.

New method:
- (At model load): For all `Ueberhoehungslinie`, find all `TOP_Kante` this bank line applies to by searching for a path between `IDUeberhoehungA` and `IDUeberhoehungB`. These values are stored in a HashMap for fast lookups.
- For each signal: Check if the `TOP_Kante` the signal is on has a `Ueberhoehungslinie`. If yes, check if the `Ueberhoehungslinie` contains the signal. If yes, calculate the bank value.
- If no value was found, search *on the same* `TOP_Kante` to either side and perform a linear calculation.


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskg/SskgTransformator.xtend
```diff
@@ -334,12 +334,12 @@ class SskgTransformator extends AbstractPlanPro2TableModelTransformator {
 	def double getShortestPathLength(Punkt_Objekt p1,
 		Punkt_Objekt p2) {
 		val points1 = p1.singlePoints.map[new TopPoint(it)]
-		val points2 = p2.singlePoints.map[new TopPoint(it)]
+		val points2 = p2.singlePoints.map[new TopPoint(it)]
```
TruongQuangSB, line 337:
```suggestion
		val points2 = p2.singlePoints.map[new TopPoint(it)]
```

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sslr/SslrTransformator.xtend
```diff
@@ -292,7 +292,7 @@ class SslrTransformator extends AbstractPlanPro2TableModelTransformator {
 		]?.filterNull ?: Collections.emptyList
 		fillSwitch(
 			cols.getColumn(Bemerkung),
-			fstrZugRangier,
+			fstrZugRangier,
```
TruongQuangSB, line 295:
```suggestion
			fstrZugRangier,
```
whitespace

TruongQuangSB (changes requested):

TruongQuangSB (approved):

PR #371, opened 2024-02-06
Use extra profile for bump version

TruongQuangSB:
- Create extra profile for set new version to avoid loading dependencises while bump version
- Now give two profile:
  - default-profile: will active by default
  - setVersion-profile: will active when call property _setVersion_


pstuecker (changes requested):
This requires additional changes in https://github.com/eclipse-set/build

pstuecker (approved):

PR #466, opened 2024-03-21
Enable browser debugging in development mode and show attachment in full size

TruongQuangSB:


java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/WebBrowser.java
```diff
@@ -43,6 +43,11 @@ public class WebBrowser {
 	 *            The SWT parent object
 	 */
 	public WebBrowser(final Composite parent) {
+		// Enable browser debug in development mode
+		if (ToolboxConfiguration.isDevelopmentMode()) {
+			System.setProperty("org.eclipse.set.browser.remote-debugging-port", //$NON-NLS-1$
+					"9999"); //$NON-NLS-1$
```
pstuecker, line 49:
```suggestion
			ChromiumStatic.getCEFConfiguration().DebugPort = 9999;
```

pstuecker (changes requested):

PR #569, opened 2024-05-13
Fix cell comparator not being transitive

pstuecker:
Java comparators must be transitive. 

Previously our cell comparator worked the following way:

1. Attempt to convert both values to ints and compare those
2. Compare the text representation of both values

However that leads to the comparator not being transitive as required by sorting algorithms. For example consider 

```
57 < 5100 (numeric)
56AA > 5100 (text)
57 > 56AA (text)
```
This leads to:
```
56AA < 57 < 5100 < 56AA
```

Instead we attempt to compare a numeric prefix. If the numeric prefixes are equal, attempt to match the remaining text (if any).

### DIFF
--- a/java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/sorting/AbstractCellComparator.xtend
+++ b/java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/sorting/AbstractCellComparator.xtend
@@ -114,10 +114,7 @@ package abstract class AbstractCellComparator implements Comparator<TableCell> {
 		val first = result.get(0)
 		// Either entirely numeric or at most 1 character text
 		if (result.length == 1) {
-			if (isInteger(first))
-				return result.get(0) -> ""
-			else
-				return "" -> result.get(0)
+			return isInteger(first) ? result.get(0) -> "" :  "" -> result.get(0)
 		}
 		// Length: 2 - Either entirely text or mixed
 		val second = result.get(1)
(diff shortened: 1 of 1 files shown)


TruongQuangSB (approved):

java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/sorting/AbstractCellComparator.xtend
```diff
@@ -83,30 +84,80 @@ package abstract class AbstractCellComparator implements Comparator<TableCell> {
...
+			return "" -> ""
+
+		val result = text.split("((?=\\D)|(?<=\\D))", 2)
+		val first = result.get(0)
+		// Either entirely numeric or at most 1 character text
+		if (result.length == 1) {
+			if (isInteger(first))
+				return result.get(0) -> ""
+			else
+				return "" -> result.get(0)
```
TruongQuangSB, line 120:
```suggestion
			return isInteger(first) ? result.get(0) -> "" :  "" -> result.get(0)
```
It is better with ternary condition

java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/sorting/AbstractCellComparator.xtend
```diff
@@ -83,30 +84,80 @@ package abstract class AbstractCellComparator implements Comparator<TableCell> {
...
+		}
+		// Length: 2 - Either entirely text or mixed
+		val second = result.get(1)
+		if (isInteger(first))
+			return first -> second
+		else
+			return "" -> first + second
```
TruongQuangSB, line 127:
```suggestion
		return isInteger(first) ? first -> second : "" -> first + second
```

TruongQuangSB (changes requested):

TruongQuangSB (approved):

PR #614, opened 2024-05-24
Implementation Jump from table to siteplan function

TruongQuangSB:
- Implementation Jump function from table (plaz, validate, planpro table) to siteplan.
- It is work like jump to source with right click on row, then selection menu item.
- When the select object doesn't guid reference, then the menu item will disable.
- If the selected object not exist in siteplan, then the menu item will disable. By not open siteplan, this check will not execute


web/siteplan/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "siteplan",
   "version": "0.1.0",
   "private": true,
+  "type": "module",
```
TruongQuangSB, line 5: The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated

web/siteplan/package.json
```diff
@@ -34,7 +35,8 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^6.1.1",
     "eslint-plugin-vue": "^9.26.0",
+    "tslib": "^2.6.2",
```
TruongQuangSB, line 38: Resolve lint warning

java/bundles/org.eclipse.set.application/OSGI-INF/l10n/bundle.properties
```diff
@@ -39,4 +39,5 @@ WebNewsPartDescriptionService_ViewName=Neuigkeiten
 ChromiumCreditsPartDescriptionService_ViewName=Chromium Lizenzinformationen
 WebDeveloperHelpPartDescriptionService_ViewName=SET-Hilfe für Entwickler
 
-TableMenuService_TextView=Zeige in Textsicht
\ No newline at end of file
+TableMenuService_TextView=Zeige in Textsicht
+TableMenuService_Siteplan=Zeige in Siteplan
```
pstuecker, line 43:
```suggestion
TableMenuService_Siteplan=Zeige in Lageplan
```

pstuecker (changes requested):

pstuecker (approved):

PR #630, opened 2024-05-30
Fix incorrect string comparisions

pstuecker:


TruongQuangSB:
should we use _equals_ function instead of _==_ operator ?

pstuecker:
In xtend == is the equals function. 
=== in Xtend is == in Java

PR #631, opened 2024-05-30
Improve error messages and infos in table overview

mariusheine:
The table overview (in debug mode) now looks like

image


java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/TableDataProvider.xtend
```diff
@@ -92,11 +93,11 @@ class TableDataProvider implements IDataProvider {
...
 
 	/** 
-	 * @param row the current table row (after filters are applied)
-	 * @return the original row index(before filters were applied)
+	 * @param row the current table row index (after filters are applied)
+	 * @return the original row index (before filters were applied)
 	 */
 	def int getOriginalRowIndex(int row) {
-		return tableContents.get(row).row.rowIndex
+		return tableContents.get(row).row.trueRowIndex
```
TruongQuangSB, line 100: It will return wrong index of row, when the table change by filter or sort. Because the table, which reference for the row isn't change by filter and sort.

TruongQuangSB (changes requested):

java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/TableDataProvider.xtend
mariusheine: @TruongQuangSB But how did it worked then before, because the classical `rowIndex` is (almost) always 0 because it gives you the index inside the rowgroup and most of the groups only have one row.

I will test it in the `PlazModelTableView`. Or is there a better place to test it?

TruongQuangSB: do you have tested with sort and filter ? When i tested with both of them, then jump to wrong number

mariusheine: @TruongQuangSB Works for me also with a filter...

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/overview/TableErrorTableColumns.xtend
```diff
@@ -30,17 +32,21 @@ class TableErrorTableColumns extends AbstractColumns {
...
 	}
 
 	def ColumnDescriptor fillHeaderDescriptions(
 		ColumnDescriptorModelBuilder builder) {
 		val GroupBuilder root = builder.createRootColumn()
-		root.add(Source).widthPercent(10)
+		root.add(Source).widthPercent(5)
+		root.add(TableType).widthPercent(5)
+		root.add(RowNumber).widthPercent(5)
 		root.add(LeadingObject).widthPercent(20)
```
TruongQuangSB, line 48:
```suggestion
		root.add(Source).widthPercent(5)
		root.add(TableType).widthPercent(10)
		root.add(RowNumber).widthPercent(5)
		root.add(LeadingObject).widthPercent(15)
```

TruongQuangSB (approved):

PR #666, opened 2024-06-12
Improve table loading popup 

TruongQuangSB:
- Avoid each Table create own pop up, when table view update

### DIFF
--- a/java/bundles/org.eclipse.set.model.titlebox/src/org/eclipse/set/model/titlebox/PlanningOffice.java
+++ b/java/bundles/org.eclipse.set.model.titlebox/src/org/eclipse/set/model/titlebox/PlanningOffice.java
@@ -1,39 +1,31 @@
 /**
- * Copyright (c) 2024 DB InfraGO AG and others
- *
- * This program and the accompanying materials are made available under the
- * terms of the Eclipse Public License v2.0 which is available at
- * https://www.eclipse.org/legal/epl-2.0.
- *
- * SPDX-License-Identifier: EPL-2.0
- *
+ * Copyright (c) {Jahr} DB Netz AG and others.
+ * 
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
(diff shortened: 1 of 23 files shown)


java/bundles/org.eclipse.set.model.titlebox/src/org/eclipse/set/model/titlebox/PlanningOffice.java
mariusheine, line 1: @TruongQuangSB The changes in this file seems to be unrelated to me. Where does it come from?

java/bundles/org.eclipse.set.model.titlebox/src/org/eclipse/set/model/titlebox/StringField.java
mariusheine, line 1: @TruongQuangSB Here as well

mariusheine (commented):
@TruongQuangSB There are a lot of other files where unrelated changes are inside. I did not have a look in all these files. Can we maybe move them in a separate PR with the context `clean up` or something like this.

The changes affecting the table service and table views (so the actual feature) looks pretty nice to me.

mariusheine:
@TruongQuangSB And can you please fix the typo in the PR title ;)

TruongQuangSB:
> @TruongQuangSB There are a lot of other files where unrelated changes are inside. I did not have a look in all these files. Can we maybe move them in a separate PR with the context `clean up` or something like this.
> 
> The changes affecting the table service and table views (so the actual feature) looks pretty nice to me.

@mariusheine ok, i have move the checkstyle fix to another PR

mariusheine:
> > @TruongQuangSB There are a lot of other files where unrelated changes are inside. I did not have a look in all these files. Can we maybe move them in a separate PR with the context `clean up` or something like this.
> > The changes affecting the table service and table views (so the actual feature) looks pretty nice to me.
> 
> @mariusheine ok, i have move the checkstyle fix to another PR

Perfect :+1:

PR #676, opened 2024-06-18
Plaz check for existence of Bearbeitungsvermerke when enum sonstige anywhere

mariusheine:


TruongQuangSB:
I think, u can filter _BasisAttribut_AttributeGroup_ by _allContents_, then find function _getWert_ with method _declaredMethods_ from Class and u can check, if the _Wert_ is _Enumerator_ or not. By that u don't need a List from BasisObject, which contains Enumerator

TruongQuangSB:
In addition we prefer to implement in Java instead of Xtext because we want to gradually move away from Xtext

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/SonstigeRequiresBearbeitungsvermerk.xtend
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Validates that any enum value "Sonstige" has an according bearbeitungsvermerk anywhere
+ * 
+ * @author mariusheine
+ */
+@Component
+class SonstigeRequiresBearbeitungsvermerk implements PlazCheck {	
+	final Check<Basis_Objekt>[] RELEVANT_ENTITIES = #[
+		new Check(Aussenelementansteuerung, [ aussenElementAnsteuerungArt ], "aussenelementansteuerungArt"),
```
pstuecker, line 35: Listing all possible types with enumerations here seems like a rather tedious task.

My suggestion would be to iterate all elements (eAllContents) and then filter for entries with a value that has an `eClass` that is an `EEnum`. That way we can identify every enumeration value in the planning file.

For those set to a Sonstige value, we can then go back up the EMF parent chain until we find the containing 
`Ur_Objekt` and check it for a remark.

pstuecker (changes requested):

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/SonstigeRequiresBearbeitungsvermerk.xtend
mariusheine: @pstuecker This does not work. the `eClass` is never an instance of or equal to `EEnum`.

I try it now with the idea of @TruongQuangSB  https://github.com/eclipse-set/set/pull/676#issuecomment-2175843950

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/SonstigeRequiresBearbeitungsvermerk.xtend
```diff
@@ -0,0 +1,102 @@
...
+		new Check(Uebertragungsweg, [ bandbreite ], "aussenelementansteuerungArt"),
+		new Check(Bahnsteig_Zugang, [ bahnsteigZugangArt ], "bahnsteigZugangArt")
+	];
+	
+	override run(IModelSession modelSession) {
+		return RELEVANT_ENTITIES.flatMap([
+			return it.singleRun(modelSession.planProSchnittstelle.eAllContents)
```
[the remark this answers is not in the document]
mariusheine: Changed it java class so this doesn't matter anymore

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/SonstigeRequiresBearbeitungsvermerk.xtend
mariusheine: This works now nicely with a combined way

mariusheine:
@pstuecker I also tried to achieve this grouping of errors like for the "Teilbereichsgrenzen" but was not able to achieve this. Maybe you can give me a hint on how to do this

pstuecker:
It actually already works, it just doesn't show up in PHausen, as groups are only created if all fields (excluding row number, line number and message) are equal, and there are at least 5 elements in the group. In PHausen all groups are of size 4 or less.

pstuecker (approved):

PR #679, opened 2024-06-18
Fix checkstyle warning

TruongQuangSB:


mariusheine:
@TruongQuangSB In #666 there have been reorderings of getters and setters in a few classes. So this is fine now to keep them in the actual order

TruongQuangSB:
> @TruongQuangSB In #666 there have been reorderings of getters and setters in a few classes. So this is fine now to keep them in the actual order

@mariusheine yes, that was automatisch reordering from Eclipse IDE 😄

PR #700, opened 2024-06-21
updated plaz model reference

mariusheine:


mariusheine:
@TruongQuangSB The changes looks fine. I verified it against the changes over the past 4 months and everything looks fine. The only thing I am bit unsure about is the `Werkzeukofferversion` as I have generated the file via elipse. Is that fine or shall I taken an "official" executable? (I forgot where to exactly find it then :see_no_evil: )

pstuecker (approved):

pstuecker:
The "official" version would be one from a maven build, as only those contain a proper version identifier.

PR #701, opened 2024-06-21
OverviewTable: add jump to table menu item

TruongQuangSB:


mariusheine:
@TruongQuangSB Everything else looked fine and worked nicely. So from my point of view you can merge it

PR #702, opened 2024-06-21
PlaZ: Implementation check for given coordinate of Punkt_Objekt

TruongQuangSB:
- In this PR:
  - Move TrackService from Siteplan to GeoKanteGeometryService
  - Move GEOKanteMetaData, GEOKanteCoordinate, GEOKanteSegment, GeoPosition to _org.eclipse.set.basis.geometry_
  - Compare the given coordinate for Punkt_Objekt through ID_GEO_Punkt_Berechnet with the calculate coordinate through siteplan


java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/GeoCoordinateValid.java
```diff
@@ -0,0 +1,167 @@
...
+		punktObjekts.forEach(po -> po.getPunktObjektTOPKante().forEach(potk -> {
+			final GEOKanteCoordinate geoKanteCoordinate = geometryService
+					.getCoordinateAt(potk, 0);
+			final Coordinate coordinate = geoKanteCoordinate.getCoordinate()
+					.getCoordinate();
+			final List<GEO_Punkt> givenGEOPunkts = potk.getIDGEOPunktBerechnet()
+					.stream().map(e -> e.getValue())
+					.filter(gp -> getNullableObject(gp,
+							e -> e.getGEOPunktAllg().getGEOKoordinatensystem()
+									.getWert()).orElse(null).equals(
```
mariusheine, line 84: @TruongQuangSB This will crash if `getGEOKoordinatensystem().getWert()`  returns `null`. Either we filter away all those that are `null` or we must do special handling for the null case or twist the object to check (but only if `geoKanteCoordinate.getCRS()` can never be null.

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/GeoCoordinateValid.java
```diff
@@ -0,0 +1,167 @@
...
+											geoKanteCoordinate.getCRS()))
+					.toList();
+			givenGEOPunkts.forEach(gp -> {
+				final Coordinate gpCoordinate = GeoPunktExtensions
+						.getCoordinate(gp);
+				if (gpCoordinate == null) {
+					result.add(creatErrorReport(gp));
+				} else if (coordinate.distance(gpCoordinate) > TOLERANT) {
```
mariusheine, line 92: @TruongQuangSB So you are checking all the coordinates of the `GeoPunktBerechnet` against the same `coordinate` from the `geoKante`. This looks a bit weird to me because in which scenario do I have multiple almost equal coordinates at a `PunktObjectTopKante`?

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/GeoCoordinateValid.java
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	@Override
+	public String getDescription() {
+		return "Alle vorgegeben GeoKoordinaten von GEO_Punkt sind plausibilisiert";
+	}
+
+	@Override
+	public String getGeneralErrMsg() {
+		return "Die vorgegebene GeoKoordinate: {GUID} weich mehr als 10cm von der berechneten Koordinate ab.";
```
mariusheine, line 154: @TruongQuangSB I would prefer to use the `TOLERANCE` value instead of hard code `10cm`. Can be simply done by `TOLERANCE * 100`

Additionally a tiny german grammar issue. It must be `weicht` instead `weich`

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/GeoCoordinateValid.java
```diff
@@ -0,0 +1,167 @@
...
+			final GEO_Punkt gp) {
+		final PlazError plazError = PlazFactory.eINSTANCE.createPlazError();
+		plazError.setSeverity(ValidationSeverity.ERROR);
+		final ID_GEO_Punkt_ohne_Proxy_TypeClass errorObject = potk
+				.getIDGEOPunktBerechnet().stream()
+				.filter(idgp -> idgp.getValue().equals(gp)).findFirst()
+				.orElse(null);
+		if (errorObject == null) {
+			throw new IllegalArgumentException(String.format(
+					"GEO_Punkt: %s ist nicht zu Punkt_Objekt: %s gehören",
```
mariusheine, line 131: @TruongQuangSB This message should be

```suggestion
					"GEO_Punkt: %s gehört nicht zu Punkt_Objekt: %s",
```

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/service/PlazModelServiceImpl.xtend
```diff
@@ -60,6 +62,9 @@ class PlazModelServiceImpl implements PlazModelService {
 
 		val errors = <PlazCheck, List<PlazError>>newHashMap
 		plazChecks.forEach [
+			if (ToolboxConfiguration.isDevelopmentMode && it instanceof GeoCoordinateValid) {
```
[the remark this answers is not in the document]
mariusheine: @TruongQuangSB While testing I just recognized that it should be

```suggestion
			if (!ToolboxConfiguration.isDevelopmentMode && it instanceof GeoCoordinateValid) {
```

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/GeoCoordinateValid.java
mariusheine: @TruongQuangSB Can we maybe integrate the actual distance between the coordinates inside the error message? Or is this breaking the grouping then?

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/GeoCoordinateValid.java
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	/**
+	 * @param potk
+	 *            the {@link Punkt_Objekt_TOP_Kante_AttributeGroup}
+	 * @return true, when exist predetermined coordinates through
+	 *         {@link Punkt_Objekt_TOP_Kante_AttributeGroup#getIDGEOPunktBerechnet}
+	 *         and the coordinates haven't same coordinate system
+	 */
+	public static boolean isNotDistincCoordinateSystem(
```
mariusheine, line 208: @TruongQuangSB Typo in the function name

```suggestion
	public static boolean isNotDistinctCoordinateSystem(
```

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/GeoCoordinateValid.java
```diff
@@ -0,0 +1,223 @@
...
+			return List.of(createProcessingWarning());
+		}
+
+		final List<PlazError> result = new ArrayList<>();
+		final List<Punkt_Objekt> punktObjekts = StreamSupport
+				.stream(container.getAllContents().spliterator(), false)
+				.filter(Punkt_Objekt.class::isInstance)
+				.map(Punkt_Objekt.class::cast).toList();
+		punktObjekts.forEach(po -> po.getPunktObjektTOPKante().forEach(potk -> {
+			if (isNotDistincCoordinateSystem(potk)) {
```
mariusheine, line 87:
```suggestion
			if (isNotDistinctCoordinateSystem(potk)) {
```

java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/transform/TrackTransformator.xtend
```diff
@@ -282,4 +284,57 @@ class TrackTransformator extends BaseTransformator<TOP_Kante> {
...
+		for (Coordinate coordinate : geoKante.geometry.coordinates) {
+			if (lastCoordinate === null) {
+				// The first (and last) coordinate is never strictly within a
+				// segment
+				// so it does not get added to the resulting list
+				lastCoordinate = coordinate
+			} else if (distance < segment.start) {
+				distance += lastCoordinate.distance(coordinate)
+				lastCoordinate = coordinate;
+			}if (distance <= segment.end) {
```
mariusheine, line 323: @TruongQuangSB But this should be 

```suggestion
			} else if (distance <= segment.end) {
```

java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/transform/TrackTransformator.xtend
```diff
@@ -282,4 +284,56 @@ class TrackTransformator extends BaseTransformator<TOP_Kante> {
...
+		for (Coordinate coordinate : geoKante.geometry.coordinates) {
+			if (lastCoordinate === null) {
+				// The first (and last) coordinate is never strictly within a
+				// segment
+				// so it does not get added to the resulting list
+				lastCoordinate = coordinate
+			} else if (distance < segment.start) {
+				distance += lastCoordinate.distance(coordinate)
+			} else if (distance < segment.end) {
+				if (result.empty) {
```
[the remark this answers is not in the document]
mariusheine: @TruongQuangSB Ok, we could solve this by extracting the for loop into an own function and make a return in that situation. Then we are safe that it is the same result.

mariusheine (approved):

java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/transform/TrackTransformator.xtend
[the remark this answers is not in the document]
TruongQuangSB: @mariusheine do u mean recusive? We can also use _switch_ insteadof _if_ . Here is example code, what do u think ?
```
		var Coordinate lastCoordinate = null
		var distance = geoKante.getStart();
		for (Coordinate coordinate : geoKante.geometry.coordinates) {
			switch (true) {
				// Initialize the previous coordinate
				case lastCoordinate === null:
					// The first (and last) coordinate is never strictly within a
					// segment
					// so it does not get added to the resulting list
					lastCoordinate = coordinate
				case distance < segment.start: {
					// Skip coordinates before the segmen
					distance += lastCoordinate.distance(coordinate)
					lastCoordinate = coordinate
				}
				case distance > segment.end:
					// Abort if we have processed all coordinates within the segment
					return result
				default: {
					if (result.isEmpty) {
						result.add(
							new GEOKanteCoordinate(lastCoordinate,
								segment.bereichObjekte, geoKante.geoKnoten.CRS))
					}
					result.add(
						new GEOKanteCoordinate(coordinate,
							segment.bereichObjekte, geoKante.geoKnoten.CRS))
					distance += lastCoordinate.distance(coordinate)
					lastCoordinate = coordinate
				}
			}
		}
```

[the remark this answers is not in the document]
mariusheine: @TruongQuangSB Would also be fine, although the `switch (true) {` looks a bit odd to me.

But if we can directly return in the case `distance > sement.end` then we can directly make return as well in the current implementation:

```java
                       } else if (distance < segment.start) {
				distance += lastCoordinate.distance(coordinate)
			} else if (distance > segment.end) {
                                return result;
                        } else {
				if (result.empty) {
```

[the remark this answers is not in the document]
TruongQuangSB: ok

mariusheine (approved):

PR #706, opened 2024-06-24
PlaZ-Model: update CRS check

TruongQuangSB:
In this PR:
 - Move multiply CRS check to separate class
 - Show reference GEO_Punkt by each invalid CRS


mariusheine (approved):
@TruongQuangSB All looks good to me.

PR #709, opened 2024-06-26
Update error message for BankValue check

TruongQuangSB:


java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/BankValues.xtend
```diff
@@ -37,38 +40,67 @@ class BankValues extends AbstractPlazContainerCheck implements PlazCheck, EventH
...
+				"Es konnte kein passender topologischer Pfad (gesucht bis {PFAD})für die Überhöhungslinie {GUID} mit ensprechend Länge ({UEBERHOEHUNG}) gefunden werden.",
+				Map.of(
+					"GUID",
+					it.identitaet?.wert,
+					"PFAD",
+					longestPath?.length?.toString,
+					"UEBERHOEHUNG",
+					bankLength.toString
+				)
+			)
```
mariusheine, line 103: @TruongQuangSB Shouldn't we return `null` here because we already return that we did not found a path on top where we check if `paths` is empty.

mariusheine (commented):
@TruongQuangSB The whole routing algorithm was hard to review but it looks correct to me. Despite this one comment for the bank values

java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/BankValues.xtend
TruongQuangSB: They are two different error:
 - First case is we can find the Path, but the length of this Path is longer than Bankinglength
 - Second case is the process to find the path was break, because it can't find relevant path with the given Banking length. With this case we will show warning with content, that the process was break and the founded path.

TruongQuangSB: @mariusheine this check will be only run, when the path can't found by BankingService:
`return container.getUeberhoehungslinie.filter [
			bankService.findTOPBanking(it) === null
		].map [`

mariusheine: @TruongQuangSB Ahh yeah. Then everything is fine.

mariusheine (approved):

PR #712, opened 2024-06-26
Siteplan: update distance from FMAComponent to track

TruongQuangSB:


mariusheine:
@TruongQuangSB This was simple :D 

Just for the future: Is there any global place where it is configured or ducmented that 1 unit in the SVG equals to `0.1mm`?

mariusheine (approved):

TruongQuangSB:
> @TruongQuangSB This was simple :D
> 
> Just for the future: Is there any global place where it is configured or ducmented that 1 unit in the SVG equals to `0.1mm`?

y, we have a document for Symbol mass

PR #717, opened 2024-06-28
Ssks: no render second line of a Signal, when this line is empty

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTransformator.xtend
```diff
@@ -944,46 +947,35 @@ class .simpleName»: «e.message» - failed to transform table contents''', e)
...
-
-			result.add(hauptbefestigungen)
-			result.add(nebenbefestigungen)
-		} else if (befestigungen.size > 2) {
-			throw new IllegalArgumentException('''«signal.bezeichnung?.bezeichnungAussenanlage?.toString» has more than two Befestigung Signal''')
-		} else {
-			result.add(rahmen.map[signalBefestigung])
-			result.add(new LinkedList<Signal_Befestigung>())
+		switch mast : befestigungen.filter [
+			mastTypeOfSignalWithTwoMast.contains(
```
mariusheine, line 959: @TruongQuangSB Previously it was only filtering by 2 enum values but now you are filtering for 6. Why?

mariusheine:
@TruongQuangSB And the reference table must be updated

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTransformator.xtend
TruongQuangSB: @mariusheine  i don't no why here filter only 2 ENUM. But through ppmtab should be more

mariusheine: @TruongQuangSB Alright. THen everything is fine

mariusheine (approved):

PR #728, opened 2024-07-03
Layout is required

mariusheine:


mariusheine:
@TruongQuangSB @pstuecker This is my initial attempt to check for the existence of `layout.xml`. Is this sufficient or do I need to check if there is an entry inside the `content.xml` as well?

TruongQuangSB:
> @TruongQuangSB @pstuecker This is my initial attempt to check for the existence of `layout.xml`. Is this sufficient or do I need to check if there is an entry inside the `content.xml` as well?

it is ok so. They are separate.

java/bundles/org.eclipse.set.feature.validation/src/org/eclipse/set/feature/validation/LayoutInfoRequired.java
```diff
@@ -0,0 +1,58 @@
...
+public class LayoutInfoRequired extends AbstractCustomValidator {
+
+	@Override
+	public void validate(final ToolboxFile toolboxFile,
+			final ValidationResult result) {
+		try {
+			toolboxFile.openLayout();
+			final PlanProFileResource layoutResource = toolboxFile
+					.getLayoutResource();
+			if (layoutResource == null) {
```
TruongQuangSB, line 34: u can get  the Layoutresource direct from Modelsession

TruongQuangSB:
@mariusheine  otherweise it should create layout.xml by new project initialize

java/bundles/org.eclipse.set.feature.validation/src/org/eclipse/set/feature/validation/LayoutInfoRequired.java
mariusheine: @TruongQuangSB But then I would need the model session inside the validation. For logic encapsulation I would actually to prefer to just work with the parameters provided from outside. 

We could think of add the model session in the AbstractCustomProvider so that all validations can assume the same context. What do you think?

TruongQuangSB: @mariusheine hmm, ur right. it is good so, my bad 😄

pstuecker (approved):

PR #751, opened 2024-07-19
Improve stability for invalid planning data

pstuecker:
- If a geometry cannot be found for a single edge, we should not abort the entire geometry run
- If a Signal_Befestigung is mounted to itself, we previously ran into an infinite loop


java/bundles/org.eclipse.set.releng.set.product/set.product
```diff
@@ -93,6 +93,7 @@
       <property name="toolbox.siteplan.export_dpi" value="600" />
       <property name="toolbox.siteplan.lod_scale" value="10000" />
       <property name="toolbox.bank_line_offset_limit" value="0.1" />
+      <property name="toolbox.siteplan.default_collisions_enabled" value="false" />
```
mariusheine, line 96: @pstuecker Is this safe to set it to false. Why is `true` by default? Can't we just set the default to `false`?

mariusheine (approved):

java/bundles/org.eclipse.set.releng.set.product/set.product
mariusheine: Will be moved into the data modell that's why it is totally fine for now

PR #752, opened 2024-07-19
Fix pdf table style not using padding for last row

pstuecker:
When we recently adjusted table styling, we did not adjust the special case for the last row on a page. This just copies the same style.

### DIFF
--- a/java/bundles/org.eclipse.set.feature/rootdir/data/export/pdf/common.xsl
+++ b/java/bundles/org.eclipse.set.feature/rootdir/data/export/pdf/common.xsl
@@ -114,6 +114,9 @@ http://www.eclipse.org/legal/epl-v20.html
 			<xsl:value-of select="$wide-border-style" />
 		</xsl:attribute>
 		<xsl:attribute name="text-align">left</xsl:attribute>
+		<xsl:attribute name="padding-left">1pt</xsl:attribute>
+		<xsl:attribute name="padding-right">1pt</xsl:attribute>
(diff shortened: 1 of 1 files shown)


mariusheine (approved):

java/bundles/org.eclipse.set.feature/rootdir/data/export/pdf/common.xsl
```diff
@@ -114,6 +114,9 @@ http://www.eclipse.org/legal/epl-v20.html
 			<xsl:value-of select="$wide-border-style" />
 		</xsl:attribute>
 		<xsl:attribute name="text-align">left</xsl:attribute>
+		<xsl:attribute name="padding-left">1pt</xsl:attribute>
+		<xsl:attribute name="padding-right">1pt</xsl:attribute>
+		<xsl:attribute name="padding-top">1pt</xsl:attribute>
```
mariusheine, line 119: @pstuecker Where exactly am I supposed to see this? In the excel or pdf export?

pstuecker: This is all for the pdf export.

mariusheine: @pstuecker OK then it doesn't work. There is no change between main and this branch in the pdf export.

mariusheine: Works when putting common.xsl at correct location in eclipse

PR #753, opened 2024-07-19
Add index column to table overview

pstuecker:


java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/overview/TableErrorTableTransformator.xtend
```diff
@@ -14,31 +14,41 @@ import org.eclipse.set.utils.table.AbstractTableModelTransformator
...
 		this.columns = columns;
 		this.enumTranslationService = enumTranslationService;
 	}
-	
-	override transformTableContent(Collection<TableError> errors, TMFactory factory) {
-		for (error : errors) {
+
+	override transformTableContent(Collection<TableError> errors,
+		TMFactory factory) {
+		val Comparator<TableError> comparator = Comparator.comparing [ TableError it |
```
pstuecker, line 32: We need to sort here, to ensure the index is applied in the correct order

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/overview/TableErrorTransformationService.java
```diff
@@ -68,9 +68,10 @@ public ColumnDescriptor fillHeaderDescriptions(
 	@Override
 	public Comparator<RowGroup> getRowGroupComparator() {
 		// default comparator
-		return TableRowGroupComparator.builder().sort("A", LEXICOGRAPHICAL, ASC) //$NON-NLS-1$
+		return TableRowGroupComparator.builder().sort("A", NUMERIC, ASC) //$NON-NLS-1$
```
mariusheine, line 71: @pstuecker When you add this index column to the sorter we must not initally sort at all because the index is always unique. 

Aditionally: Is the grouping of rows working nevertheless? It doesn't work from my point of view

mariusheine (approved):

mariusheine (approved):

PR #778, opened 2024-07-31
Modify SWTBot

TruongQuangSB:
- The test file will open only onetimes by table data test instead of reopen per table
- Extract get test resource path to own function


mariusheine (approved):
@TruongQuangSB This of course as well. Forgot to click approve 😄

PR #792, opened 2024-08-05
Add zero width space for line break

TruongQuangSB:
**Titlebox by field _Bauzustand Zeichen_**:
 - Add _zero width space_ to long text without space  to automatic line break

### DIFF
--- a/java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/utils/PlanProToTitleboxTransformation.xtend
+++ b/java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/utils/PlanProToTitleboxTransformation.xtend
@@ -49,6 +49,8 @@ class PlanProToTitleboxTransformation {
 	 */
 	public static final String DATE_FORMAT_SHORT = "MM/yy"; // $NON-NLS-1$
 	static final String EMPTY_PLANNING_NUMBER = "<Plannummer DB AG>"
+	static final int SMALL_FELD_MAX_LENGTH = 9
+	static final String ZERO_WIDTH_SPACE = "\u200b"
 
 	private new() {
 	}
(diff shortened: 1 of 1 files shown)


java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/utils/PlanProToTitleboxTransformation.xtend
```diff
@@ -192,8 +194,13 @@ class PlanProToTitleboxTransformation {
...
 	private def String getBauzustandPlanungAllgemein(
 		Planung_E_Allg_AttributeGroup planung) {
-		return String.format("%s", // $NON-NLS-1$
-		planung?.getBauzustandKurzbezeichnung()?.getWert() ?: "");
+			val bauzustand = planung?.getBauzustandKurzbezeichnung()?.getWert() ?: ""
+			if (bauzustand.length > SMALL_FELD_MAX_LENGTH) {
+				val sb = new StringBuilder(bauzustand)
+				sb.insert(SMALL_FELD_MAX_LENGTH, ZERO_WIDTH_SPACE)
+				return sb.toString
+			}
```
mariusheine, line 202: @TruongQuangSB I have added the following value for the `BauzustandKurzbezeichnung`: `wwwwwwwwwwwwwwwwwwwwwww` 

This led to the following issues:
- 9 characters for the max length is insufficient because 9 times `w` is already too wide
- if external tools are sending more than 15 letters we need a third line

As a solution I would suggest to do:
bauzustand.split(' ').map(word -> // insert ZERO_WIDTH_SPACE at every 7th character )

mariusheine: image

mariusheine (changes requested):

java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/utils/PlanProToTitleboxTransformation.xtend
TruongQuangSB: ok, i have change max length text to 7. The text shouldn't have more than 15 character. We have a pattern for check this

mariusheine: @TruongQuangSB Perfect

mariusheine (approved):

mariusheine (approved):

PR #798, opened 2024-08-06
Export table through shortcut

TruongQuangSB:
- Add Export function through shortcut (like PT1) table to PlaZ-Model and Validation table.
- SWTBot: Export current and reference data of Plaz-Model and Validation table, when test failed


mariusheine (approved):
@TruongQuangSB Looks fine to me

PR #799, opened 2024-08-06
Fix exception by save action

TruongQuangSB:
Fix exception:
 - When change data in _Projekt Bearbeiten_ and save to new file by loaded Project without _layout_ content.

### DIFF
--- a/java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/fileservice/PlainFileFormatService.java
+++ b/java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/fileservice/PlainFileFormatService.java
@@ -40,8 +40,8 @@ public class PlainFileFormatService implements ToolboxFileFormatService {
 
 	@Override
 	public ToolboxFile create(final Format format, final ToolboxFileRole role) {
-		return new PlainToolboxFile(sessionService, format,
-				createEditingDomain());
+		throw new UnsupportedOperationException(
+				"It isn't supported creat new model als plain format"); //$NON-NLS-1$
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/fileservice/PlainFileFormatService.java
```diff
@@ -40,8 +40,8 @@ public class PlainFileFormatService implements ToolboxFileFormatService {
 
 	@Override
 	public ToolboxFile create(final Format format, final ToolboxFileRole role) {
-		return new PlainToolboxFile(sessionService, format,
-				createEditingDomain());
+		throw new UnsupportedOperationException(
+				"It isn't supported creat new model als plain format"); //$NON-NLS-1$
```
TruongQuangSB, line 44: At the moment it can only new project initiliaze for *.planpro format

java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/fileservice/ZippedPlanProToolboxFile.java
```diff
@@ -108,10 +127,14 @@ private static File newFile(final Path dir, final ZipEntry entry)
 		this.loadable = loadable;
 		this.temporaryDirectory = null;
 		this.role = role;
-		setResource(TECHNICAL_RESOURCE_TYPE_NAME, createPlanProResource());
-		setResource(LAYOUT_RESOURCE_TYPE_NAME, createPlanProResource());
```
TruongQuangSB, line 112: We don't set empty resource, when loading file

java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/fileservice/PlainFileFormatService.java
mariusheine, line 44: @TruongQuangSB Just some typo 

```suggestion
				"It isn't supported to create a new model in plain format"); //$NON-NLS-1$
```

mariusheine (approved):
@TruongQuangSB I will see if it works fine when I test it later. From code perspective it looks pretty unrelated for now. But I think this is due lack of knowledge of file loading and saving at my side.

PR #886, opened 2024-09-19
Fix exception and add null check

TruongQuangSB:


mariusheine (approved):
@TruongQuangSB Only one small typo

mariusheine (approved):

PR #906, opened 2024-09-27
Compare View for Footnotes

TruongQuangSB:


mariusheine (commented):
@TruongQuangSB Where are all these icons actually used? I did not saw them in the app. This whole package `org.eclipse.set.model.tablemodel.edit` does not make sense to me 😕 . Why is it necessary?

mariusheine:
> > @TruongQuangSB Where are all these icons actually used? I did not saw them in the app. This whole package `org.eclipse.set.model.tablemodel.edit` does not make sense to me 😕 . Why is it necessary?
> 
> all package with org.eclipse.set.model.tablemodel.* is generate from tablemodel.ecore. It is not require to check 😄

@TruongQuangSB Good because I did not checked them 🤣

java/bundles/org.eclipse.set.feature/rootdir/data/export/pdf/common.xsl
```diff
@@ -231,10 +231,32 @@ http://www.eclipse.org/legal/epl-v20.html
...
+
+	<xsl:template match="OLD_FOOTNOTE">
+		<fo:block text-align="left">
+			<fo:inline background-color="yellow" text-decoration="line-through">
+		*				<xsl:value-of select="@footnote-number" />
+:				<xsl:value-of select="." />
 
+			</fo:inline>
+		</fo:block>
+	</xsl:template>
```
mariusheine, line 261: @TruongQuangSB This is not working in my tests. I don't see the footnote numbers anymore in the PDF and the coloring is also not done. I tried to fix it but without any luck.

mariusheine: @TruongQuangSB Yes now it is working fine 😄

mariusheine (approved):

PR #1026, opened 2024-10-29
Ssza: Update content of coulm F and I

TruongQuangSB:


mariusheine (approved):
@TruongQuangSB Looks fine, but there are still the other two cases where we also do `coordinate.distance(...`. Shouldn't we also fix them or is there something else different?

TruongQuangSB:
> @TruongQuangSB Looks fine, but there are still the other two cases where we also do `coordinate.distance(...`. Shouldn't we also fix them or is there something else different?

@mariusheine u are righe, i overlooked it

mariusheine (approved):

PR #1056, opened 2024-11-11
Improve jump to siteplan

mariusheine:
Selecting a feature sets the lod scale to the default value. (1:1000 at the moment)

Additonally 
- fixed a small typo 
- added a helper script for easier development between SET main application and siteplan browser app.

### DIFF
--- a/java/bundles/org.eclipse.set.releng.set.product/set.product
+++ b/java/bundles/org.eclipse.set.releng.set.product/set.product
@@ -91,7 +91,7 @@
       <property name="toolbox.siteplan.sheetcut_crs" value="DR0" />
       <property name="toolbox.table.tree.minimum" value="5" />
       <property name="toolbox.siteplan.export_dpi" value="600" />
-      <property name="toolbox.siteplan.lod_scale" value="10000" />
+      <property name="toolbox.siteplan.lod_scale" value="1000" />
       <property name="toolbox.bank_line_offset_limit" value="0.1" />
       <property name="toolbox.siteplan.default_collisions_enabled" value="false" />
    </configurations>


web/siteplan/.env.example
TruongQuangSB, line 1: @mariusheine we can simple comment in eclipse-dev-build.sh oder package.json, aren't ?

web/siteplan/src/components/MapContainer.vue
```diff
@@ -76,7 +76,7 @@ export default class MapContainer extends Vue {
     // Allow clicking the scale bar to force a 1:1000 scale
     if (store.state.planproModelType === PlanProModelType.SITEPLAN) {
       this.scaleLine = new ScaleBarControl(() => {
-        this.scaleLocked = 1000
+        this.scaleLocked = Configuration.getLodScale()
```
TruongQuangSB, line 79: @mariusheine pls, change the default LOD-Scale in configuration.json and in eclipse product. Currently the default is set to 10.000

TruongQuangSB (changes requested):

web/siteplan/.env.example
mariusheine: @TruongQuangSB Yes we could but if anyone of use puts the eclipse folder somewhere else he/she will always have changes in its git tree. That's why I put it as env variable so that it is installation independent.

TruongQuangSB: ok

web/siteplan/src/components/MapContainer.vue
TruongQuangSB: @mariusheine miss here org.eclipse.set.releng.set.product/set.product

mariusheine: @TruongQuangSB Changed

TruongQuangSB (approved):

PR #1078, opened 2024-11-20
linting improvements

mariusheine:
- the `java-format.sh` is there to fix formatting on the entire codebase locally
- checkstyle job is not failing because no check is marked as error. They are all warnings
- removed project specific code formatter configurations


mariusheine:
> > @TruongQuangSB Please do a review here. I was a bit unsure if I should fix all formatting issues in here as well or we should do it in another PR. I fixed them for now but we can easily revert that commit if you are of a different opinion
> 
> @mariusheine i think, we should fix formatting in another PR. It make better overview

@TruongQuangSB Alright. I have reverted it. I will make a second PR after this one is merged.

mariusheine:
@TruongQuangSB I also updated build instructions and removed remaining project specific configurations of the java code formatter

java-format.sh
TruongQuangSB, line 1: @mariusheine do we need this batch? Is it for local format, isn't it ?

.github/workflows/quality.yml
```diff
@@ -0,0 +1,80 @@
...
+  push:
+    tags:
+      - 'v**'
+    branches:
+      - 'release/**'
+      - 'main'
+
+
+jobs:
+  checkstyle:
```
TruongQuangSB, line 14: @mariusheine do we need checkstyle job here ? Because by build Java Job is checkstyle already executed

java-format.sh
mariusheine: @TruongQuangSB Yes it is just for local formatting so that you don't need to open all the files and format them by yourself. But we don't need it. I can remove it if you want

.github/workflows/quality.yml
mariusheine: @TruongQuangSB Ahh yeah you are right. In my opinion it creates more awareness as a separate job. In the maven build logs it simply get lost. SO we actually don't need it.

java-format.sh
TruongQuangSB: @mariusheine yeah, but this should only one times execute, isn't it. After the Format consistent, then shouldn't more Format error

.github/workflows/quality.yml
TruongQuangSB: @mariusheine Maven checkstyle in `Build SET` workflow isn't active 😄. Can u pls active

java-format.sh
mariusheine: @TruongQuangSB I have removed the script

.github/workflows/quality.yml
mariusheine: I activated it

TruongQuangSB (approved):

PR #1083, opened 2024-11-22
Ssks: Fix distance between neighbor track

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTransformator.xtend
mariusheine, line 1: @TruongQuangSB So the error was "just" in the wrong angle of the signal? Because all the other logic seems similar to me.

Ok and the case with wirkrichtung_beide was not there before.

mariusheine (approved):

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTransformator.xtend
TruongQuangSB: @mariusheine it is wrong rotation angle of the perpendicular. For `wirkrichtung_beide` should the angle like `wirkrichtung_in`

mariusheine:
@TruongQuangSB Looking at the table differences for the Ssks https://github.com/eclipse-set/set/issues/1124#issuecomment-2514715314 a lot of distances switched direction. Is this still correct then?

TruongQuangSB:
> @TruongQuangSB Looking at the table differences for the Ssks [#1124 (comment)](https://github.com/eclipse-set/set/issues/1124#issuecomment-2514715314) a lot of distances switched direction. Is this still correct then?

i have analyse with TextView and looks change fine. We have here with direction of Signal Mast to defintion left/right side distance. But i give a strange, that is the direction of the mast and the signal aren't same 😄 . I think, we must first with Carsten declare

mariusheine:
> > @TruongQuangSB Looking at the table differences for the Ssks [#1124 (comment)](https://github.com/eclipse-set/set/issues/1124#issuecomment-2514715314) a lot of distances switched direction. Is this still correct then?
> 
> i have analyse with TextView and looks change fine. We have here with direction of Signal Mast to defintion left/right side distance. But i give a strange, that is the direction of the mast and the signal aren't same 😄 . I think, we must first with Carsten declare

@TruongQuangSB Alright. Will you clarify it with Carsten or shall we do it together?

TruongQuangSB:
> > > @TruongQuangSB Looking at the table differences for the Ssks [#1124 (comment)](https://github.com/eclipse-set/set/issues/1124#issuecomment-2514715314) a lot of distances switched direction. Is this still correct then?
> > 
> > 
> > i have analyse with TextView and looks change fine. We have here with direction of Signal Mast to defintion left/right side distance. But i give a strange, that is the direction of the mast and the signal aren't same 😄 . I think, we must first with Carsten declare
> 
> @TruongQuangSB Alright. Will you clarify it with Carsten or shall we do it together?

After speak with Carsten, the `beide` direction is a illegal direction.  Therefore it will be a lot `WARNING` symbol now :smile:

java/bundles/org.eclipse.set.nattable.utils/src/org/eclipse/set/nattable/utils/PlanProTableCellPainter.java
```diff
@@ -30,27 +29,27 @@ public class PlanProTableCellPainter extends BackgroundPainter {
...
 						new WarningPainter(blackWarningImage,
 								CellContentExtensions.WARNING_MARK_BLACK)),
-						BaseEdgeEnum.BEFORE,
-						new WarningPainter(leftWarningImage,
-								CellContentExtensions.WARNING_MARK_YELLOW)),
 				BaseEdgeEnum.AFTER,
-				new WarningPainter(rightWarningImage,
-						CellContentExtensions.WARNING_MARK_RED)),
+				new WarningPainter(yellowWarningImage,
+						CellContentExtensions.WARNING_MARK_YELLOW)),
```
TruongQuangSB, line 52: Resort old/new value.
 1. New value
 2. Common value
 3. Old value

java/bundles/org.eclipse.set.nattable.utils/src/org/eclipse/set/nattable/utils/BasePainterDecorator.java
```diff
@@ -90,7 +90,7 @@ public Rectangle getDecoratorCellPainterBounds(final ILayerCell cell,
 			case RIGHT:
 				return new Rectangle(adjustedCellBounds.x,
 						adjustedCellBounds.y + preferredBaseHeight
-								- preferredDecoratorHeight,
+								- getSpacing(),
```
TruongQuangSB, line 93: fix warning icon overlapp text

mariusheine (approved):

PR #1139, opened 2024-12-09
Fix jump to text view function by signal

TruongQuangSB:


web/siteplan/src/components/popup/MenuPopup.vue
```diff
@@ -160,14 +161,15 @@ export default class MenuPopup extends Vue {
         featureType !== FeatureType.Flash
     }).flatMap(ele => {
       const featureType = getFeatureType(ele)
-      const featureData = getFeatureData(ele)
+      const featureData = getFeatureData(ele) as SignalMount
```
mariusheine, line 164: @TruongQuangSB Thi type cast should not be here in my opinion because featureData can still be something else here. And I would say that it is also not necessary because we are casting it in line 166

mariusheine (changes requested):

web/siteplan/src/components/popup/MenuPopup.vue
TruongQuangSB: u a right

TruongQuangSB, line 164:
```suggestion
      const featureData = getFeatureData(ele)
```

mariusheine (approved):

PR #1171, opened 2024-12-13
Ssks: Reconstructed determine signal side distance

TruongQuangSB:


java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/ToolboxConfiguration.java
```diff
@@ -132,8 +132,9 @@ public static String getShortName() {
 	 * @return whether the toolbox is in development mode
 	 */
 	public static boolean isDevelopmentMode() {
-		return Boolean.parseBoolean(System.getProperty(
-				ToolboxProperties.DEVELOPMENT_MODE, Boolean.TRUE.toString()));
+		// return Boolean.parseBoolean(System.getProperty(
+		// ToolboxProperties.DEVELOPMENT_MODE, Boolean.TRUE.toString()));
+		return false;
```
mariusheine, line 137: @TruongQuangSB I think this should be reverted, shouldn't it?

mariusheine (changes requested):

java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/ToolboxConfiguration.java
TruongQuangSB: ja genau 😄

PR #1179, opened 2024-12-16
Fix missing railway by development mode

TruongQuangSB:


web/siteplan/src/feature/TrackFeature.ts
```diff
@@ -292,38 +293,88 @@ export default class TrackFeature extends LageplanFeature<Track> {
...
     })
-
-    return isDiffDesignation || isDiffSection
+    const test = isDiffDesignation || exsitDiffSection
       ? this.createCompareFeatures(initial, final)
       : this.getFeatures(final)
+    return test
+  }
+
+  protected createCompareFeatures (initial: SiteplanState, final: SiteplanState): Feature<Geometry>[] {
```
mariusheine, line 330: @TruongQuangSB So the actual bug was that the default compare feature calculation was not suitable for tracks, correct?

web/siteplan/src/feature/TrackFeature.ts
```diff
@@ -292,38 +293,88 @@ export default class TrackFeature extends LageplanFeature<Track> {
...
+
+        if (finalFeature) {
+          finalSegmentFeatures.forEach(feature => {
+            if (!result.includes(feature)) {
+              result.push(feature)
+            }
+          })
+        }
+
+        result.push(initialFeature)
```
mariusheine, line 365: @TruongQuangSB Do we really need to push the initialFeature if we have found a final feature? Or is the initial feature the one painted in red in the siteplan?

mariusheine (changes requested):

web/siteplan/src/feature/TrackFeature.ts
TruongQuangSB: correct

web/siteplan/src/feature/TrackFeature.ts
TruongQuangSB: when the inital and the final state aren't same, then we need add both feature to visual change. Also:
- red is final
- yellow is initial

web/siteplan/src/feature/TrackFeature.ts
```diff
@@ -292,38 +293,88 @@ export default class TrackFeature extends LageplanFeature<Track> {
...
       })
-      return isNotEqual
     })
-
-    return isDiffDesignation || isDiffSection
+    const test = isDiffDesignation || exsitDiffSection
       ? this.createCompareFeatures(initial, final)
       : this.getFeatures(final)
+    return test
+  }
```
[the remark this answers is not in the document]
TruongQuangSB: ok

web/siteplan/src/feature/TrackFeature.ts
mariusheine: @TruongQuangSB 🙈 Yes initial is yellow...

And we only find a `finalFeature` if there are differences? Or where do we checked if intial and final aren't same?

TruongQuangSB: y, the `finalFeature` will only by differences to `result` added 
It was compare in line 354, the variable name isn't correct. `finalFeautre` should be `changed` or `isEqual`

mariusheine: @TruongQuangSB Alright then it looks fine. Can you please change the variable accordingly then?

TruongQuangSB: ok

web/siteplan/src/feature/TrackFeature.ts
```diff
@@ -292,38 +293,84 @@ export default class TrackFeature extends LageplanFeature<Track> {
     ])
     const initialTracks = this.getObjectsModel(initial)
     const finalTracks = this.getObjectsModel(final)
-    const isDiffSection = initialTracks.some(initialTrack => {
+    const exsitDiffSection = initialTracks.some(initialTrack => {
```
mariusheine, line 296: @TruongQuangSB A small typo

```suggestion
    const existDiffSection = initialTracks.some(initialTrack => {
```

web/siteplan/src/feature/TrackFeature.ts
```diff
@@ -292,38 +293,84 @@ export default class TrackFeature extends LageplanFeature<Track> {
...
+          const initialSection = initialTrack.sections.find(x => x.guid === finalSection.guid)
+          if (initialSection && compare(finalSection, initialSection)) {
+            this.setObjectColor(initialTrack, finalSection.guid, SiteplanColorValue.COLOR_REMOVED)
+            this.setObjectColor(finalTrack, finalSection.guid, SiteplanColorValue.COLOR_ADDED)
+            return true
+          }
+        })
       }
+    })
+    return isDiffDesignation || exsitDiffSection
```
mariusheine, line 321:
```suggestion
    return isDiffDesignation || existDiffSection
```

mariusheine (changes requested):

mariusheine (approved):

PR #1198, opened 2024-12-19
Modify Ssld examplar and Free reporting section

TruongQuangSB:
- Fix signal side distance
- Fix free reporting section length
- Modify Ssld examplar


java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/graph/TopPath.java
```diff
@@ -72,15 +115,27 @@ public BigDecimal firstEdgeLength() {
...
 		for (final TOP_Kante edge : edges()) {
 			final BigDecimal edgeLength = edge.getTOPKanteAllg().getTOPLaenge()
 					.getWert();
 
 			// If the point is on the current edge, check which direction to add
 			if (point.edge().equals(edge)) {
+				// Point in first Edge
+				if (previousEdge == null && point.edge() == edge) {
+					// Point lie before path start
+					if (point.distance()
```
mariusheine, line 132: @TruongQuangSB What if the we need to look in inverse direction? So imagine the following:

- an edge `E` from node `A` to node `B`  with a point `P` on it
- `E` has length `10`
-  `P` has distance `7`
- the distance of `P` is the distance from node `A` on the edge `E` (correct)?
- if `startOfPath` is node `A`, then we correctly evaluate that the point is on the edge
- But if `startOfPath` is node `B`, then we get an empty result because `10` is less than `7`. What I would expect here would be `3` as a result

Or am I getting something wrong?

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssld/SsldTransformator.xtend
```diff
@@ -57,45 +58,53 @@ class SsldTransformator extends AbstractPlanPro2TableModelTransformator {
...
-		// which in direction of the start Signal
-		if (fmas.empty) {
-			fmas = dweg?.fmaAnlageFreimeldung?.map[fmaGrenzen]?.flatten.toSet.
-				filter[topGraph.isInWirkrichtungOfSignal(startSignal, it)].
-				toList
-		}
-
-		val relevantDistances = fmas?.map [
-			getShortestPathLength(dweg?.fstrFahrweg?.start, it)
+		val fmas = dweg?.fmaAnlageFreimeldung?.map[fmaGrenzen]?.flatten.toSet.
```
mariusheine, line 75: @TruongQuangSB We are taking gleisabschluesse here also into consideration. From my understanding we only must pay attention to the Achszaehlpunkte here :thinking:

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssld/SsldTransformator.xtend
```diff
@@ -57,45 +58,53 @@ class SsldTransformator extends AbstractPlanPro2TableModelTransformator {
...
+				topKante
+			]
+			val relevantPaths = pathFromSignalToFMA.filter [
+				value.edges.forall[fstrTOPKante.contains(it)]
+			]
+			if (relevantPaths.empty) {
+				throw new IllegalArgumentException("no path found")
+			}
+			return relevantPaths.maxBy[value.length].value.length.
+				toTableIntegerAgateDown
```
mariusheine, line 101: @TruongQuangSB This logic seems not totally correct. Remember the scenario `5 + B`. If Dweg 5 would extend the FMA pretty long then your logic might take `Länge B (Abzweig)` although the Dweg never takes the Abzweig.

java/bundles/org.eclipse.set.feature/rootdir/data/export/excel/ssld_vorlage.xlsx
mariusheine, line 1: @TruongQuangSB What have you changed here? I would need to download the file and compare it, so maybe you can just tell it here :stuck_out_tongue:

java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/graph/TopPath.java
```diff
@@ -40,6 +41,48 @@ public TopPath(final List<TOP_Kante> edges, final BigDecimal length,
 		this.edges = edges;
 		this.length = length;
 		this.firstEdgeLength = firstEdgeLength;
+		this.startNode = null;
```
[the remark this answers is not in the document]
TruongQuangSB: ok

java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/graph/TopPath.java
TruongQuangSB: @mariusheine u are recht, the inverse direction case wasn't considered

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssld/SsldTransformator.xtend
TruongQuangSB: @TruongQuangSB when the DWeg longer than FMA, then i find the FMA, which the path from start signal to this FMA all in Dweg ist. Then take the longest Path. The `Abzweig` path will be not considered

java/bundles/org.eclipse.set.feature/rootdir/data/export/excel/ssld_vorlage.xlsx
TruongQuangSB: @mariusheine  the column Q und R was change name

java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/graph/TopPath.java
```diff
@@ -72,49 +113,76 @@ public BigDecimal firstEdgeLength() {
...
+			return Optional.of(BigDecimal.ZERO);
+		}
+
+		final boolean inTopDirection = edgeLength.subtract(startNode.distance())
+				.compareTo(firstEdgeLength) == 0;
+		// Point lie before path start
+		if (inTopDirection == point.distance()
+				.compareTo(startNode.distance()) < 0) {
+			return Optional.empty();
+		}
```
mariusheine, line 183: @TruongQuangSB The equal 0 case makes me struggle a bit when not in top direction. I think we should simply do

```suggestion
		if (
		        (inTopDirection && point.distance().compareTo(startNode.distance()) < 0) ||
		        (!inTopDirection && point.distance().compareTo(startNode.distance()) > 0)
		) {
			return Optional.empty();
		}
```

java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/graph/TopPath.java
```diff
@@ -72,49 +113,76 @@ public BigDecimal firstEdgeLength() {
...
+				.subtract(point.distance());
+		if (pointDistance.compareTo(BigDecimal.ZERO) < 0
+				|| pointDistance.compareTo(length) > 0) {
+			return Optional.empty();
+		}
+
+		return Optional.of(pointDistance);
+	}
+
+	private Optional<BigDecimal> pointOnFirstEdge(final TopPoint point,
```
mariusheine, line 171: @TruongQuangSB I would prefer to name it `getDistanceOnFirstEdge`

web/siteplan/src/components/popup/TrackSectionPopup.vue
mariusheine, line 1: @TruongQuangSB The changes here just help for orientation and "debugging", correct? Or are there any deeper requirements related?

mariusheine (changes requested):

web/siteplan/src/components/popup/TrackSectionPopup.vue
TruongQuangSB: @mariusheine this ist just for test. it will be revert

java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/graph/TopPath.java
TruongQuangSB: ok

java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/graph/TopPath.java
mariusheine: @TruongQuangSB Ahh yes you are right. Have overseen this :smile:

mariusheine (approved):

PR #1206, opened 2024-12-20
Corrector EKW and DKW visualization

TruongQuangSB:


java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/transform/SiteplanTransformatorImpl.xtend
```diff
@@ -64,7 +65,13 @@ class SiteplanTransformatorImpl extends AbstractSiteplanTransformator {
 		siteplan.transformLayout(modelSession.layoutInformation)
 		siteplan.eAllContents.filter(SiteplanObject).filter[guid !== null].
 			forEach [
-				siteplanService.addSiteplanElement(it)
+				if (it instanceof TrackSwitch) {
+					components.forEach [ componente |
+						siteplanService.addSiteplanElement(it)
```
mariusheine, line 70: @TruongQuangSB So here you are adding the track switch multiple times. Sure that we don't need to add the `componente`? Why do we the `forEach` otherwise?

TruongQuangSB: @mariusheine normaly by simple TrackSwitch have a component, but by crossing switch like DWK, EKW the trackswitch have 2 component. In siteplan, javascript part, we created for each component a Feature. That why, i have add each component here

mariusheine: @TruongQuangSB Yeah of course, but you are calling `siteplanService.addSiteplanElement(it)` where `is` is the `TrackSwitch` and not the `component`. So are you calling two times `siteplanService.addSiteplanElement(trackSwitch)` for DKW, EKW and so on.

TruongQuangSB: ah, u are right 😄

mariusheine (approved):

mariusheine (approved):

PR #1207, opened 2024-12-20
Sszw: fill value by column L and M

TruongQuangSB:


mariusheine (changes requested):

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sszw/SszwColumns.java
```diff
@@ -76,7 +76,7 @@ public class SszwColumns {
 	/**
 	 * Sszw.Ansteuerung.ESTW_Zentraleinheit
 	 */
-	public static final String ESTW_Zentraleinheit = "L";
+	public static final String ESTWZentraleinheit = "L";
```
TruongQuangSB, line 79: rename to different with Class ESTW_Zentraleinheit

PR #1223, opened 2025-01-06
Handle Station label undefined

TruongQuangSB:


web/siteplan/src/feature/StationFeature.ts
```diff
@@ -60,9 +60,10 @@ export default class StationFeature extends LageplanFeature<Station> {
         textRotation += 180
       }
 
+      const text = station.label?.text ? 'Bstg ' + station.label?.text : '⚠'
       return new Style({
         text: new Text({
-          text: 'Bstg ' + station.label?.text,
+          text,
```
mariusheine, line 66: @TruongQuangSB Since the station label is optional we should not show anything

```
      if (!station.label) {
        return new Style();
      }
      return new Style({
        text: new Text({
          text: 'Bstg ' + station.label.text,

mariusheine (changes requested):

web/siteplan/src/feature/StationFeature.ts
TruongQuangSB: ok

mariusheine (approved):

PR #1239, opened 2025-01-10
Fix find node by line number

TruongQuangSB:
- ValidationReport:
  - Fix nodefinder with line number
  - Modify find ObjectScope and Object Attribute name

### DIFF
--- a/java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/xml/NodeAttributeNameProvider.java
+++ b/java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/xml/NodeAttributeNameProvider.java
@@ -19,6 +19,7 @@ import org.w3c.dom.NodeList;
  */
 public class NodeAttributeNameProvider {
 	private static final String TEXT_NODE = "#text"; //$NON-NLS-1$
+	private static final String DOCUMENT_NODE = "#document"; //$NON-NLS-1$
 	private static final String CONTAINER_NODE = "Container"; //$NON-NLS-1$
 
 	private static final String IDENTITY_ATTRIBUTE_NAME = "Identitaet"; //$NON-NLS-1$
@@ -30,9 +31,11 @@ public class NodeAttributeNameProvider {
 	 * @return The closest XML attribute name (or null)
 	 */
 	public String getAttributeName(final Node node) {
-		// When the parent node is container node, then the node isn't
+		// When the parent node is container node or root node, then the node
+		// isn't
 		// attribute/group node
-		if (node == null
+		if (node == null || node.getParentNode() == null
+				|| node.getParentNode().getNodeName().equals(DOCUMENT_NODE)
 				|| node.getParentNode().getNodeName().equals(CONTAINER_NODE)) {
 			return null;
 		}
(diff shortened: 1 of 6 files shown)


java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/xml/NodeAttributeNameProvider.java
```diff
@@ -30,9 +31,11 @@ public class NodeAttributeNameProvider {
 	 * @return The closest XML attribute name (or null)
 	 */
 	public String getAttributeName(final Node node) {
-		// When the parent node is container node, then the node isn't
+		// When the parent node is container node or root node, then the node
+		// isn't
 		// attribute/group node
-		if (node == null
+		if (node == null || node.getParentNode() == null
+				|| node.getParentNode().getNodeName().equals(DOCUMENT_NODE)
```
TruongQuangSB, line 38: root node haven't attribute name

java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/xml/NodeObjectScopeProvider.java
```diff
@@ -105,6 +113,13 @@ private boolean isGUIDInLSTPlanungsbereich(final Node node,
 		}
 
 		return guidPlanungsbereichCache.contains(guid);
+	}
+
+	private boolean isInContainerNode(final Node node) {
+		if (node == null || node.getNodeName().equals(NODE_CONTAINER)) {
```
mariusheine, line 119: @TruongQuangSB Why are we retaining `false` when the node name is `Container`? From the function name I would have expect `true` to be returned.

mariusheine (changes requested):

java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/xml/NodeObjectScopeProvider.java
TruongQuangSB: @mariusheine i have here forget condition for returned `true`. When the node is container node, then it isn't in container :D

java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/xml/NodeObjectScopeProvider.java
```diff
@@ -105,6 +113,13 @@ private boolean isGUIDInLSTPlanungsbereich(final Node node,
...
+			return false;
+		}
 
+		return node.getParentNode().equals(NODE_CONTAINER) || isInContainerNode(node.getParentNode());
```
mariusheine, line 123: @TruongQuangSB Can it be now that `node.getParent()` is null and **not** the container node? Maybe we should catch this nevertheless

TruongQuangSB: @mariusheine  ah yes. I have modify. THis will return true, when node name Equals CONTAINER_NODE

mariusheine (approved):

PR #1259, opened 2025-01-20
[siteplan] show red yellow comparism also in non-developer mode

mariusheine:


web/siteplan/src/util/SVG/Draw/SvgDraw.ts
```diff
@@ -94,6 +95,12 @@ export default class SvgDraw {
     labelsvg.setAttribute('id', 'label')
     labelsvg.setAttribute('font-size', fontSize.toString())
     labelsvg.setAttribute('font-family', 'siteplanfont')
+    if (label.color && applyColor) {
+      const rgbColorString = `rgb(${label.color[0]}, ${label.color[1]}, ${label.color[2]})`
+      labelsvg.setAttribute('fill', rgbColorString)
+      labelsvg.setAttribute('stroke', rgbColorString)
+    }
```
TruongQuangSB, line 102: @mariusheine i think, we don't need set label color here. We already use svg `filter` for set svg color (see SvgColorService).

TruongQuangSB (changes requested):

web/siteplan/src/util/SVG/Draw/SvgDraw.ts
mariusheine: @TruongQuangSB Yeah but it somehow was not applied for the signal km marker

java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/SiteplanBrowser.java
```diff
@@ -131,6 +143,16 @@ private void serveSiteplan(final Response response)
...
+		if (ToolboxConfiguration.isDevelopmentMode()) {
+			try {
+				final Path path = Paths.get("siteplan.json");
+				Files.writeString(path, json);
+				System.out.println("Exported currently loaded siteplan.json to "
+						+ path.toAbsolutePath());
+			} catch (final IOException e) {
+				// don't care for now. It is just development mode stuff
+			}
+		}
```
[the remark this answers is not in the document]
mariusheine: @TruongQuangSB Ahh damn. The export function makes this obsolete :rofl:

web/siteplan/src/components/FeatureService.vue
```diff
@@ -230,40 +228,32 @@ export default class FeatureService extends Vue {
...
-            .flat()
-            .map(feature => featureClass.setFeatureColor(feature))
+          const compareState = featureClass.compareChangedState(
+            model.changedInitialState,
+            model.changedFinalState
+          )
+          return [
+            featureClass
+              .getFeatures(model.commonState)
+              .map(feature => featureClass.setFeatureColor(feature, SiteplanColorValue.COLOR_UNCHANGED_PLANNING)),
```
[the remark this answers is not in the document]
mariusheine: @TruongQuangSB But then it is probably also wrong to set the color to `UNCHANGED_PLANNING` in initial and final state, isn't it?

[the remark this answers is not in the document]
TruongQuangSB: @mariusheine yes, in initial and final state shouldn't set color to `UNCAHNGED_PLANNING`

web/siteplan/src/util/SVG/Draw/SvgDraw.ts
TruongQuangSB: @mariusheine the reason is the Label not contain `ObjectColor`. Because of this can't find the `objectColor` property in `SvgService #getSVGStyle` line 94. You can change the propery `color` in Label to `objectColors` with type ObjectColor. It will be work

mariusheine: @TruongQuangSB This did not worked totally, because the SvgService tried to get the objectColors from the drawData (which was the `attachedSignals[0]` of the `signalMount`). So I had to set the correct drawData (the `signalMount` in this case) and it worked nicely.

TruongQuangSB: @mariusheine the `drawData` isn't only `attachedSignals[0]`, it is contain `attachedSignals[0]`, `featureType`, `label`. But it ist fine, when you change to `signalMount`, By SingalRouteMarker doesn't matter `attachedSignals[0]` or `signalMount`

java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/SiteplanBrowser.java
```diff
@@ -105,6 +106,17 @@ private void serveConfiguration(final Response response)
...
+			try {
+				final Path path = Paths.get("configuration.json");
+				Files.writeString(path, json);
+				System.out.println(
+						"Exported currently loaded configuration.json to "
+								+ path.toAbsolutePath());
+			} catch (final IOException e) {
+				// don't care for now. It is just development mode stuff
+			}
+		}
```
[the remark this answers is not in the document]
mariusheine: Reverted it

java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/SiteplanBrowser.java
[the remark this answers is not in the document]
mariusheine: Reverted it

web/siteplan/src/model/Signal.ts
```diff
@@ -60,7 +60,7 @@ export enum SignalPart {
     Schirm = 'schirm',
     Mast = 'mast',
     Additive = 'additive',
-    RouteMarker = 'marker'
+    RouteMarker = 'label'
```
TruongQuangSB, line 63: @mariusheine here shouldn't still `marker`. The signal label and marke are different

mariusheine: @TruongQuangSB Ok let me try to find a different solution. Without that change it was not easily possible to set the correct color on the marker

mariusheine: @TruongQuangSB And the reason why it is not working is because we are statically assigning the `id="label"` to the signal route marker in `SvgDraw#getRouteMarker`. Any idea how to fix this?

mariusheine: @TruongQuangSB By thinking about this I came across that I did not set the correct color on the signal label 🙈 Let me check further

TruongQuangSB: @mariusheine u can simple Add enum `Label` to SignalPart and modify SignalFeature#setFeatureRegionColor to colored the label

mariusheine: @TruongQuangSB Nice that i exactly what I test at the moment. Seems to work without any issue

mariusheine: @TruongQuangSB Pushed it. Please review it one last time

web/siteplan/src/util/SVG/Draw/SvgDraw.ts
```diff
@@ -256,7 +256,9 @@ export default class SvgDraw {
       13,
       90,
       true
-    ))
+    )
+    labelSvg.setAttribute('id', 'marker')
```
TruongQuangSB, line 260: @mariusheine use enum instead fix string

mariusheine: @TruongQuangSB Good idea. Done!

TruongQuangSB (approved):

PR #1273, opened 2025-01-22
Sskp: fix missing table entry

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskp/SskpTransformator.xtend
```diff
@@ -121,91 +122,91 @@ class SskpTransformator extends AbstractPlanPro2TableModelTransformator {
...
-			instance,
-			cols.getColumn(PZB_Schutzpunkt),
-			dweg,
-			[isPZB2000],
-			[
-				IDPZBGefahrpunkt?.value?.bezeichnung?.
-					bezeichnungMarkanterPunkt?.wert
-			]
-		)
+		if (isPZB2000 && dweg !== null) {
```
mariusheine, line 125: @TruongQuangSB Why have you switched the pattern to that "global" instead of using `fillConditional`?

TruongQuangSB: i think, this is simpler and avoid more time check

mariusheine (approved):

PR #1285, opened 2025-01-27
Ssks: update sort piority

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTransformationService.java
```diff
@@ -91,4 +99,13 @@ public void handleEvent(final Event event) {
...
 		}
 	}
+
+	@SuppressWarnings("nls")
+	@Override
+	public Comparator<RowGroup> getRowGroupComparator() {
+		return TableRowGroupComparator.builder()
+				.sort("B", LEXICOGRAPHICAL_EMPTY_LAST, ASC)
+				.sort("D", LEXICOGRAPHICAL_EMPTY_LAST, ASC)
+				.sort("A", LEXICOGRAPHICAL, ASC).build();
```
mariusheine, line 109: @TruongQuangSB Shouldn't we just sort by column `A`? At least including `B` feels strange because it depends on the sorting inside the column itself...

mariusheine (approved):

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTransformationService.java
mariusheine: Nevermind. I was in the wrong table 🙈

PR #1297, opened 2025-01-28
Ssks: specifier pdf export

TruongQuangSB:
- Make specifier pdf export for Table Ssks
- Bei Ssks Table: the fiktiv and real Signal will be sperate Pdf page render


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTablePdfExportBuilder.java
```diff
@@ -0,0 +1,205 @@
...
+import org.w3c.dom.Document;
+import org.xml.sax.SAXException;
+
+/**
+ * Specifier PDF export for table Ssks via FOP
+ * 
+ * @author Truong
+ */
+@Component(immediate = true, service = TableExport.class)
+public class SsksTablePdfExportBuilder extends FopPdfExportBuilder {
```
TruongQuangSB, line 69: This class have some duplicate code with `FopPdfExportBuilder`. THe duplicate should be move to a Abstractclass.

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTablePdfExportBuilder.java
```diff
@@ -0,0 +1,224 @@
...
+		final Optional<Entry<Integer, TableRow>> lastFiktiveSignalEntry = fiktiveSignalRows
+				.stream().max((first, second) -> Integer.compare(
+						first.getKey().intValue(), second.getKey().intValue()));
+		if (firstFiktiveSignalEntry.isPresent()) {
+			result.add(firstFiktiveSignalEntry.get().getKey().toString());
+		}
+
+		if (lastFiktiveSignalEntry.isPresent() && lastFiktiveSignalEntry.get()
+				.getKey().intValue() < rows.size()) {
+			// To break a Pdf page with XSL, the attribute "break-bfore" was
```
mariusheine, line 199: @TruongQuangSB Here you are adding a pagebreak after the last fiktive signal in case we have more signals, correct? I guess this page break is not added when don't have any fiktive signal. Maybe we should add a page break then as well?

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTablePdfExportBuilder.java
```diff
@@ -0,0 +1,224 @@
...
+						Paths.get(outputDir, getFilename(shortcut, "xml")), //$NON-NLS-1$
+						tableDocumentText);
+			}
+			createSsksTablePdf(table, tableDocumentText, outputPath, shortcut,
+					tableType, PdfAMode.PDF_A_3a, overwriteHandling);
+		} catch (final ParserConfigurationException | TransformerException
+				| IOException | SAXException e) {
+			throw new FileExportException(outputPath, e);
+		}
+	}
```
mariusheine, line 124: @TruongQuangSB Instead of overwriting this whole export method, we could introduce a parameter `pageBreaks` or `subBlocks` in the `FopPdfExportBuilder` which can then simply passed to `super.export(..., this.getPageBreakRowsIndex(table))`. There we could also integrate the info for the text field.

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTablePdfExportBuilder.java
mariusheine: I don't think it is necessary to add a new abstract class for this. We just need to restructure something in the `FopPdfExportBuilder`. See also my comment below

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTablePdfExportBuilder.java
TruongQuangSB: @mariusheine y, when we have more another Signal type, then should be in separate page. You are right, it will not break, when no fiktiv signal give. I will make that break at last of real and fiktiv Signal. And `break-after` will be used instead of `break-before`, then we don't need +1 more

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTablePdfExportBuilder.java
TruongQuangSB: @mariusheine y, good idea

java/bundles/org.eclipse.set.services/src/org/eclipse/set/services/export/TableExport.java
```diff
@@ -94,4 +101,24 @@ void exportTitleboxImage(Titlebox titlebox, Path imagePath,
...
+	 * @return export format of this service
+	 */
+	ExportFormat getExportFormat();
+
+	/**
+	 * @param table
+	 *            the export table
+	 * @return the index of row, which the export should break page
+	 */
+	List<String> getPageBreakRowsIndex(final Table table);
```
mariusheine, line 123: @TruongQuangSB One last thing: I think we don't need that method for all exports. Why not keeping this method local for the `FopPdfExportBuilder`. Then `SsksTablePdfExportBuilder` can still override it but everything else stays clean.

TruongQuangSB: ok

mariusheine (approved):

PR #1335, opened 2025-02-10
Siteplan: Fix EKW visualation

TruongQuangSB:


web/siteplan/vite.config.js
```diff
@@ -1,7 +1,11 @@
 import vue from '@vitejs/plugin-vue'
+import dotenv from 'dotenv'
 import path from 'path'
 import {defineConfig} from 'vite'
 
+dotenv.config({
+  path: './config.env'
```
mariusheine, line 7: @TruongQuangSB Is there any particular reason why you took `config.env` instead of the default `.env`?

mariusheine: @TruongQuangSB Ok so it means we are taking default `.env`? We maybe should apply this approach to the other web packages as well. See https://github.com/eclipse-set/set/pull/1345#issuecomment-2664779588

mariusheine: Yes great

web/siteplan/vite.config.js
mariusheine, line 1: @TruongQuangSB But in my local setup I still need to call `dotenv.config()` otherwise it does not automatically load my `.env` file

mariusheine (approved):

PR #1341, opened 2025-02-12
Fix missing banking by Ssks table

TruongQuangSB:


java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/graph/TopPath.java
```diff
@@ -78,17 +98,50 @@ private TopPoint determineStartNode() {
...
+					edges.get(1));
+			final BigDecimal connectTopKnotenDistance = connectTopKnoten == firstEdge
+					.getIDTOPKnotenA()
+					.getValue() ? BigDecimal.ZERO : edgeLength;
+			// When the start node lie at TOP_Knoten_A or TOP_Knoten_B, then the
+			// first edge length is ZERO
+			if (startNode.distance().compareTo(connectTopKnotenDistance) == 0) {
+				return BigDecimal.ZERO;
+			}
+			return edgeLength.subtract(startNode.distance());
```
mariusheine, line 118: @TruongQuangSB This is now pretty similar code to `determineStartNode`. Can#T we do one method return a `Pair<TopPoint, BigDecimal> determineStart` to not do the calculation twice?

mariusheine (changes requested):

java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/graph/TopPath.java
TruongQuangSB: @mariusheine It will call only one of both methoed in constructor. When we determine them in a method, then it will be more complex. Because we can only find `startNode` or `firstEdgeLength`, when on of them is already defined.

mariusheine (approved):

PR #1346, opened 2025-02-17
Update Document Export View

TruongQuangSB:


java/bundles/org.eclipse.set.feature.export/src/org/eclipse/set/feature/export/parts/PlanProExportPart.java
TruongQuangSB, line 1: Move to `org.eclipse.set.feature.table.export`

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceImpl.java
```diff
@@ -582,4 +590,26 @@ private void wireCacheSupplier() {
 		Digraphs.setEdgeToSubPathCacheSupplier(() -> getCache()
 				.getCache(ToolboxConstants.CacheId.DIRECTED_EDGE_TO_SUBPATH));
 	}
+
+	@Override
+	public Map<TableInfo, Table> transformTables(final IProgressMonitor monitor,
```
mariusheine, line 595: @TruongQuangSB Such a monitor is also/already used in the validation view where we can trigger a validation for all tables. We should reuse this there as well.

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/messages/Messages.java
```diff
@@ -121,4 +121,48 @@ public class Messages {
...
+	/**
+	 * Ausgabe des Einzelzustands der PlanPro-Datei.
+	 */
+	public String TableExportPart_ExportStateDescription;
+
+	/** Export des Dokumentensatz durchführen */
+	public String TableExportPart_ExportButton;
+
+	/**
+	 * Ohne leer Tabellen
```
mariusheine, line 154:
```suggestion
	 * Leere Tabellen abwählen
```

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceImpl.java
```diff
@@ -480,15 +486,17 @@ public Table transformToTable(final String elementId,
 
 	@Override
 	public void updateTable(final BasePart tablePart,
-			final List<String> tableCategories,
+			final List<Pt1TableCategory> tableCategories,
 			final Runnable updateTableHandler, final Runnable clearInstance) {
 		// Find which table categories should be update
 		final List<String> tablePrefixes = List
 				.of(ToolboxConstants.ESTW_TABLE_PART_ID_PREFIX,
 						ToolboxConstants.ETCS_TABLE_PART_ID_PREFIX)
```
mariusheine, line 494: @TruongQuangSB What is with the ESTW-Supplement tables? Don't we need to handle them here as well?

java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/TableInfo.java
```diff
@@ -0,0 +1,78 @@
...
+ * @param category
+ *            the table area (ESTW or ETCS)
+ * @param shortcut
+ *            the table shortcut
+ */
+public record TableInfo(Pt1TableCategory category, String shortcut) {
+	/**
+	 * The table category enum
+	 */
+	public enum Pt1TableCategory {
```
mariusheine, line 26: @TruongQuangSB Do we still need the categories then in the `ToolboxConstants`?. Isn't it safer to rely on this enum everywhere?

mariusheine (commented):
Two questions for my understanding?
1. The moving of a lot of code from `org.eclipse.set.feature.export` to `org.eclipse.set.feature.table` was necessary because we needed stuff from the tables and would have ceated cyclic dependency otherwise, correct?
2. Where does all these changes in the Manifest Files are coming from? At least a few them looked pretty unrelated to the current changes to me.

java/bundles/org.eclipse.set.feature.export/src/org/eclipse/set/feature/export/parts/PlanProExportPart.java
mariusheine: @TruongQuangSB Do you rememeber if you had to change something inside there?

java/bundles/org.eclipse.set.feature.export/src/org/eclipse/set/feature/export/checkboxmodel/CheckboxTreeModel.java
```diff
@@ -0,0 +1,170 @@
...
+		elements.forEach(ele -> ele.select());
+	}
+
+	/**
+	 * @return the parent element in tree model
+	 */
+	public CheckBoxTreeElement[] getParentElements() {
+		return Arrays.stream(getAllElements())
+				.filter(ele -> ele.isParent())
+				.toArray(CheckBoxTreeElement[]::new);
```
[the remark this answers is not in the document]
TruongQuangSB: @mariusheine this function return all Parent elements (in whole tree) and `this.elements` is only first level element

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceImpl.java
TruongQuangSB: @mariusheine u are right, here is missing ESTW-SUppelment tables

java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/export/CheckBoxTreeElement.java
```diff
@@ -0,0 +1,117 @@
...
+		this.childs.add(child);
+	}
+
+	@Override
+	public void select() {
+		super.select();
+		if (isParent()) {
+			childs.forEach(ele -> ele.select());
+		} else if (parent != null && parent.isAllChildSameState(true)) {
+			parent.selected = true;
```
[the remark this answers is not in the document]
TruongQuangSB: @mariusheine the second `if` is execute, when we `child element` selected. It will be check, if all child of `this .parent` selected or not, then change the state of `parent`.  When the selected is `parent`, then all childs (deeply child too) will be selected, this is first condition.

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceImpl.java
TruongQuangSB: @mariusheine the another function in Table overview will be use this function

java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/TableInfo.java
TruongQuangSB: @mariusheine u are right

TruongQuangSB:
> Two questions for my understanding?
> 
>     1. The moving of a lot of code from `org.eclipse.set.feature.export` to `org.eclipse.set.feature.table` was necessary because we needed stuff from the tables and would have ceated cyclic dependency otherwise, correct?
> 
>     2. Where does all these changes in the Manifest Files are coming from? At least a few them looked pretty unrelated to the current changes to me.

@mariusheine 
1. Yes and i find this more generic, because the PlanProExportPart, etc are only for export table
2. It can be automatisch resort of import list. otherweise is *.xml for osgi service

java/bundles/org.eclipse.set.feature.table/OSGI-INF/l10n/bundle.properties
```diff
@@ -22,3 +22,15 @@ TableErrorTableColumns_Source=Quelle
...
+
+TableExportPart_ExportPlanningRecordsDescriptionService_ViewName=Dokumentensatz erstellen
+TableExportPart_ExportInventoryDescriptionService_ViewName=Bestandsdokumente erstellen
+TableExportPart_Export_button=Dokumentensatz erstellen
+TableExportPart_ExportInventoryRecords=Bestandsdokumente erstellen
+TableExportPart_TaskMsg=Tabellenexport
+TableExportPart_ExportInventoryDescription=Ausgabe des Zielzustands der Planung.
+TableExportPart_ExportPlanningDescription=Ausgabe des rot/gelb Vergleichs (pdf) und des Zielzustands der Planung (xlsx).
+TableExportPart_ExportStateDescription=Ausgabe des Einzelzustands der PlanPro-Datei.
+TableExportPart_FilterEmptyButton=leere Tabelle abwählen
```
mariusheine, line 36: @TruongQuangSB Just one more correction. I was not seeing it before properly. I am sorry

```suggestion
TableExportPart_FilterEmptyButton=leere Tabellen abwählen
```

java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/TableInfo.java
mariusheine: @TruongQuangSB Can we remove the related toolbox constants now?

mariusheine (changes requested):

java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/TableInfo.java
TruongQuangSB: @mariusheine y, i have forgot 😀

mariusheine (approved):

PR #1360, opened 2025-02-20
Fix Siteplan export

TruongQuangSB:


web/siteplan/src/util/Controls/ExportControl.ts
```diff
@@ -131,7 +136,12 @@ export default class ExportControl extends Control {
         const topRightSphereCoord = transform(getTopRight(viewportExtent), 'EPSG:3857', 'EPSG:4326')
         const distanceInMeter = getDistance(topLeftSphereCoord, topRightSphereCoord)
         view.setResolution(currentResolution)
-        resolve(viewportSize[0] / distanceInMeter)
+        const cuurentPpm = viewportSize[0] / distanceInMeter
```
mariusheine, line 139:
```suggestion
        const currentPpm = viewportSize[0] / distanceInMeter
```

mariusheine:
@TruongQuangSB What was the actual problem for that nested siteplan inside that one page of the pdf? What was caused by the wrong scale?

TruongQuangSB:
> @TruongQuangSB What was the actual problem for that nested siteplan inside that one page of the pdf? What was caused by the wrong scale?

@mariusheine i don't know why, but the problem is calculate pixel per meter😅. By first time export everything is ok, but by 2. time then we meet the Problem, when we try to calculate pixel per meter again

mariusheine (approved):

PR #1451, opened 2025-03-14
Siteplan: fix compare view

TruongQuangSB:


web/siteplan/package.json
```diff
@@ -17,7 +17,6 @@
     "@turf/intersect": "^7.2.0",
     "axios": "^1.8.1",
     "geojson": "^0.5.0",
-    "jspdf": "^3.0.0",
```
TruongQuangSB, line 20: we don't need this dependency anymore

mariusheine (approved):

PR #1462, opened 2025-03-19
Cached table transform and fix element deselect

TruongQuangSB:


mariusheine:
@TruongQuangSB There was a bug when you we triggered the deselect all and had some children of a tree element already selected. Then these selected children have not been deselected. Maybe you can review my latest commit. Everything else looked and worked fine

mariusheine (approved):

PR #1493, opened 2025-04-01
Ssks change lichtraumprofile calculation

mariusheine:


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTransformator.xtend
```diff
@@ -296,14 +296,11 @@ class SsksTransformator extends AbstractPlanPro2TableModelTransformator {
...
-								val s = it
-								val mountpoints = gruppe.map [
-									s.getMountPoint(it)
-								]
-								val topEdges = mountpoints.map[topKante].toSet
-								topEdges.map [
-									gleisLichtraum?.lichtraumprofil?.wert?.
-										translate
+								val lichtraeume = it.topKanten.map [
+									gleisLichtraum
```
TruongQuangSB, line 300: @mariusheine we no need this function more. To find the `Gleis_Lichtraum` of a signal, we can use the method `org.eclipse.set.ppmodel.extensions.BereichObjektExtensions#contains`. Otherweise here should be also filter the `Gleis_Lichtraum`, which have same `lichtraumprofil`value

mariusheine: @TruongQuangSB I have changed it. Please have a look again

TruongQuangSB (approved):

PR #1504, opened 2025-04-07
Add Open Recent Menu

TruongQuangSB:


java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/configurationservice/UserConfigurationServiceImpl.java
```diff
@@ -157,6 +161,7 @@ public void addKnownVersions(final String version) throws IOException {
 	@Override
 	public void setLastFileOpenPath(final Path path) {
 		configuration.lastFileOpenPath = path;
+		getLastOpenFiles().addFirst(path);
```
mariusheine, line 164: @TruongQuangSB We should also avoid duplicates.

```suggestion
		List<Path> lastOpenFiles = getLastOpenFiles();
		if (lastOpenFiles.contains(path)) {
			// remove path if already inside open files to avoid duplicates
			lastOpenFiles.remove(path);
		}
		lastOpenFiles.addFirst(path);
```

mariusheine (changes requested):

java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/configurationservice/UserConfigurationServiceImpl.java
TruongQuangSB: here should be compare path als string

TruongQuangSB:
> @TruongQuangSB I was not able to open one of the recent projects when clicking on it in the menu.

@mariusheine  fixed

mariusheine:
> @mariusheine fixed

@TruongQuangSB Nice that works, but 🤣 when I open a project via the menu now that project is not moved to the top of recent projects. We should add this behavior there as well

TruongQuangSB:
> > @mariusheine fixed
> 
> @TruongQuangSB Nice that works, but 🤣 when I open a project via the menu now that project is not moved to the top of recent projects. We should add this behavior there as well

@mariusheine ok. i was fixed

mariusheine (approved):

PR #1511, opened 2025-04-08
Rendere ENUM 'sonstige' 

TruongQuangSB:
- Visualation `Bearbeitungsvermerk.Kurztext` instead of enum value, when the Enumerator have `sonstige` value

### DIFF
--- a/java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sslr/SslrTransformator.xtend
+++ b/java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sslr/SslrTransformator.xtend
@@ -59,8 +59,8 @@ class SslrTransformator extends AbstractPlanPro2TableModelTransformator {
 
 	private def Table create factory.table transform(
 		MultiContainer_AttributeGroup container, Stell_Bereich controlArea) {
-		container.fstrZugRangier.filter[isPlanningObject]
-			.filterObjectsInControlArea(controlArea).filter[isR].forEach [ it |
+		container.fstrZugRangier.filter[isPlanningObject].
+			filterObjectsInControlArea(controlArea).filter[isR].forEach [ it |
 				if (Thread.currentThread.interrupted) {
 					return
 				}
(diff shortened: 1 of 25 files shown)


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sslr/SslrTransformator.xtend
```diff
@@ -123,15 +123,15 @@ class SslrTransformator extends AbstractPlanPro2TableModelTransformator {
 			cols.getColumn(Art),
 			fstrZugRangier,
 			[
-				fstrRangier?.fstrRangierArt?.wert?.translate?.substring(0, 1)
+				fstrRangier?.fstrRangierArt?.translateEnum?.substring(0, 1)
```
mariusheine, line 126: @TruongQuangSB This seems falsy to me for the case when we return with the note text

mariusheine:
@TruongQuangSB If we achieve to resolve all table differences from #1515 by today I would like to wait until this before merging. If not I will merge it in the evening.

mariusheine:
> @TruongQuangSB If we achieve to resolve all table differences from #1515 by today I would like to wait until this before merging. If not I will merge it in the evening.

@TruongQuangSB Nevermind. SWT Bot is comparing against current state of the main branch and not the reference tables so we don't have any differences in PPHN. Despite my comment above everything is fine then

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sslr/SslrTransformator.xtend
TruongQuangSB: @mariusheine what do you mean ?

mariusheine: @TruongQuangSB I mean what if `fstrRangierArt.wert` is `sonstige` and it has a note, then we would run `substring(0, 1)` on that note which might not be what we and the user wants asthe first letter of the note might be equal to the first letter of any real existing `fstrRangierArt`.

TruongQuangSB: u are right

PR #1512, opened 2025-04-08
Sskw update replacement drive

TruongQuangSB:


java/bundles/org.eclipse.set.feature.export/src/org/eclipse/set/feature/export/tablediff/CustomTableDiffService.java
```diff
@@ -111,6 +114,15 @@ private static void createDiffContent(final int i, final TableRow row,
...
 			oldCell.setContent(compareContent);
+		} else if (oldCell
+				.getContent() instanceof final MultiColorCellContent content) {
+			final MultiColorCellContent newCellContent = TablemodelFactory.eINSTANCE
+					.createMultiColorCellContent();
+			final List<MultiColorContent> contents = new ArrayList<>(
+					content.getValue());
+			contents.forEach(ele -> ele.setToBeRender(true));
+			newCellContent.getValue().addAll(contents);
+			oldCell.setContent(newCellContent);
```
mariusheine, line 125: @TruongQuangSB What do we need this for? It seems unrelated to the actual feature

java/bundles/org.eclipse.set.model.tablemodel/model/tablemodel.ecore
```diff
@@ -179,6 +179,7 @@
   <eClassifiers xsi:type="ecore:EClass" name="MultiColorContent">
     <eStructuralFeatures xsi:type="ecore:EAttribute" name="multiColorValue" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString"/>
     <eStructuralFeatures xsi:type="ecore:EAttribute" name="stringFormat" eType="ecore:EDataType http://www.eclipse.org/emf/2002/Ecore#//EString"/>
+    <eStructuralFeatures xsi:type="ecore:EAttribute" name="toBeRender" eType="ecore:EDataType http://www.eclipse.org/emf/2003/XMLType#//Boolean"/>
```
mariusheine, line 182: @TruongQuangSB The name `toBeRender` is not good in my opinion. What we actually want is to disable the multi coloring in certain situations. So why not just handle it as `disableMultiColor`?

mariusheine (changes requested):

java/bundles/org.eclipse.set.feature.export/src/org/eclipse/set/feature/export/tablediff/CustomTableDiffService.java
TruongQuangSB: @mariusheine we should only by `Table-DIFF` state red/yellow visulation. I have no better idea than create a Attribute to know, when the `MutilColorCellContent` red/yellow rendere.

mariusheine (approved):

mariusheine:
@TruongQuangSB Seems like the SWT Bot can not generate the table diffs. It does not find Nattable. Can you please have a look on that?

TruongQuangSB:
@mariusheine it  work again

PR #1518, opened 2025-04-09
Sskz: Add new columns

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskz/SskzTransformator.java
```diff
@@ -177,16 +219,140 @@ private Table transform(final Iterable<Aussenelementansteuerung> controls,
...
+
+				final List<Pt1TableChangeProperties> changeProperties = new ArrayList<>();
+				waitingFillTrackMitteDistance.forEach((row, control) -> {
+					refillTrackMitteDistance(row, container, control,
+							changeProperties);
+				});
+			}
+		}, String.format("%s/%s/s", tableShortcut.toLowerCase(), //$NON-NLS-1$
+				ToolboxConstants.CacheId.GEOKANTE_GEOMETRY,
+				container.getCacheString())).start();
```
mariusheine, line 279: @TruongQuangSB We should move this thread into the `if` above, shouldn't we? Because we need a `potk` in order to calculate it properly.

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskz/SskzTransformator.java
mariusheine, line 1: @TruongQuangSB I did not compare the code against how it is implemented in the `Ssks`. Do you have done any adjustments compared to `Ssks`? And if yes, where?

TruongQuangSB: @mariusheine here have i only table name adjustments 😃

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskz/SskzTransformator.java
TruongQuangSB: @mariusheine no, because this threaad should run after all the transform done. And it will be only look at the Map `waitingFillTrackMitteDistance`,

mariusheine: @TruongQuangSB Ahh yes you are right!

PR #1523, opened 2025-04-10
Avaiable change CRS on normal mode

TruongQuangSB:


web/siteplan/src/components/development/SettingEditor.vue
```diff
@@ -15,67 +15,88 @@
       <h4>Einstellungen</h4>
       <div class="model-summary-control-content">
         <ul>
+          <div v-if="isDevelopmentMode()">
+            <li>
+              <input
+                id="checkbox-development"
+                v-model="isDevelopmentMode"
+                type="checkbox"
+              >
```
mariusheine, line 24: @TruongQuangSB I think this doesn't work because `isDevelopmentMode` is a readonly computed and not writeable one. We didn't had this before so lets just remove it again

mariusheine (changes requested):

web/siteplan/src/components/development/SettingEditor.vue
TruongQuangSB: @mariusheine y. it is a mistake

mariusheine (approved):

PR #1524, opened 2025-04-10
update GeoKanteUniqueCoordinate

TruongQuangSB:


java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/MeridianBetweenGEOKante.xtend
```diff
@@ -37,37 +38,44 @@ class MeridianBetweenGEOKante extends AbstractPlazContainerCheck implements Plaz
...
+			errList.add(err)
+
 		]
 		return errList
 	}
 
+	static def boolean isMeridianGEOKante(GEO_Kante edge) {
+		if (edge.GEOKanteAllg?.GEOLaenge?.wert.doubleValue === 0.0) {
+			return false;
+		}
```
mariusheine, line 62: @TruongQuangSB I would not put this here as a meridian geo kante can actually have length 0. Additionally
- it was not in the filter function above before this PR
- in `GeoKanteUniqueCoordinate` we are checking for kante length 0 before

TruongQuangSB: @mariusheine the meridian geo kante should have length 0. When not, then is a error. Before it was a condition in Line 60

mariusheine (approved):

PR #1527, opened 2025-04-10
only calculate lichtraum when real signal

mariusheine:


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTransformator.xtend
```diff
@@ -289,20 +289,22 @@ class SsksTransformator extends AbstractPlanPro2TableModelTransformator {
...
-								val s = it
-								val lichtraeume = it.container.gleisLichtraum.
-									filter[contains(s)]
-								lichtraeume.map [
-									lichtraumprofil?.translateEnum
-								].toSet.toList.sort
-							],
-							null
-						)
+						if (signal.signalReal !== null) {
```
TruongQuangSB, line 292: @mariusheine we can use `fillIterableWithConditional` instead of add a `if` condition

mariusheine: @TruongQuangSB Ahh yeah you are right

TruongQuangSB (approved):

PR #1529, opened 2025-04-10
selected cell highlighting

mariusheine:


java/bundles/org.eclipse.set.nattable.utils/src/org/eclipse/set/nattable/utils/PlanProTableThemeConfiguration.java
```diff
@@ -158,6 +158,9 @@ public PlanProTableThemeConfiguration(final NatTable natTable,
 				GUIHelper.COLOR_DARK_GRAY);
 		this.defaultSelectionBgColor = GUIHelper.getColor(173, 216, 230);
 		this.defaultSelectionFgColor = GUIHelper.COLOR_BLACK;
+		this.selectionAnchorSelectionBgColor = GUIHelper.getColor(173, 216,
+				230);
```
TruongQuangSB, line 162: @mariusheine you can't see selected cell hightlight more, wenn it have same color with selected row/column. I think, we should deactive seletec Cell when the row/column is selected

PR #1532, opened 2025-04-10
Sskw: fix red/yellow visulation

TruongQuangSB:


java/bundles/org.eclipse.set.feature.export/src/org/eclipse/set/feature/export/tablediff/CustomTableDiffService.java
```diff
@@ -107,25 +105,79 @@ private static void createDiffContent(final int i, final TableRow row,
...
-			contents.forEach(ele -> ele.setDisableMultiColor(true));
-			newCellContent.getValue().addAll(contents);
-			oldCell.setContent(newCellContent);
+			oldCell.setContent(createCompareCellContent(oldValue, newValue,
+					oldCell.getContent().getSeparator()));
 		}
 	}
 
+	// IMPROVE: this function isn't completely.
+	private static void createMultiColorDiffCotent(final TableCell oldCell,
```
TruongQuangSB, line 121: This is only fixed for Sskw. This function should be completely later

mariusheine: @TruongQuangSB What exactly is sskw specific?

TruongQuangSB: @mariusheine by Sskw is red/yellow/black visualiation in a Cell content. Normal can a cell value only red or yellow or black

PR #1595, opened 2025-05-13
Fix Control area condition

TruongQuangSB:


java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/BereichObjektExtensions.xtend
mariusheine, line 1: @TruongQuangSB The changes in here a bit complex. I assume there have been two bugs:

1. we have checked too late for `sameTopKante` and have returned false instead
2. The check if the point was really in the teilbereich was based on wrong begrenzungsWerte

Is that correct?

TruongQuangSB: @mariusheine the Bug is the condition `sameTopKante` never `true` 😅. The function `adjacentTopKanten` delivery only the TOP_Kante, which connected with the TOP_Kante of `teilbereich`. That mean, when the TOP_Kante of `teilbereich` is same the TOP_Kante of singlepoint, then the condition by Row 562-563 will `false`

PR #1644, opened 2025-06-04
Sszs: fix column C,D,E,P,V

TruongQuangSB:
- Fix swapped value in column C,D,E
- Fix Error by column P
- Update column V: when `-3<= value <= 5` then fill 0


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sszs/SszsTransformator.xtend
mariusheine, line 1: The formatting of the distances in column `P` is now weird. It shows the numbers with a comma as decimal separator where else in other columns it uses a dot.

And the values are left aligned. I think it should be centered

mariusheine (approved):

PR #1680, opened 2025-06-23
Improve project compare

TruongQuangSB:


java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/geometry/Chord.java
```diff
@@ -26,7 +25,7 @@ public class Chord {
 	/**
 	 * Step size for linearization
 	 */
-	private static final double STEP_SIZE = PI / 100;
+	private static final double STEP_SIZE = 0.0001;
```
mariusheine, line 28: @TruongQuangSB This change is unrelated and also causes changes in ugbb

mariusheine (changes requested):

java/bundles/org.eclipse.set.basis/src/org/eclipse/set/basis/geometry/Chord.java
TruongQuangSB: @mariusheine  oh yes 😁. This is test code

mariusheine (approved):

PR #1741, opened 2025-07-16
Fix missing B pages by Ssks export

TruongQuangSB:
- Error: By Ssks export missing complete B page
  - The reason is, the result of transform `maxPaperWidth` to String is `-39,0000000` and the value isn't valid for fop


java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/export/xsl/MultiPageTableHeader.java
```diff
@@ -77,8 +79,10 @@ protected Document getFopPageTemplate() throws ParserConfigurationException,
 				multipageTemplate, FO_BLOCK, ATTR_ID, START_INDENT);
 		if (newPageBlock.isPresent()) {
 			newPageBlock.get()
-					.setAttribute(START_INDENT, String.format("-%fcm", //$NON-NLS-1$
-							Float.valueOf(maxPaperWidth)));
+					.setAttribute(START_INDENT,
+							BigDecimalExtensions.toTableDecimal(
+									BigDecimal.valueOf(maxPaperWidth).negate(),
```
mariusheine, line 84: @TruongQuangSB What was the reason for this error. Because previously we had `String.format("-%fcm", ...` so I think the minus came due to the format.

PR #1757, opened 2025-07-23
Update control area condition

TruongQuangSB:


java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/AussenelementansteuerungExtensions.xtend
```diff
@@ -110,7 +110,24 @@ class AussenelementansteuerungExtensions extends BasisObjektExtensions {
...
+			Aussenelementansteuerung).toList
+		val areaAEAInformationPrimaer = areaAEA.informationPrimaer.filter(
+			Aussenelementansteuerung).toList
+
+		return currentInformationPrimaer.exists[it === areaAEA] ||
+			areaAEAInformationPrimaer.exists [ firstAEA |
+				currentInformationPrimaer.exists [ secondAEA |
+					secondAEA === firstAEA
+				]
+			]
```
mariusheine, line 130: @TruongQuangSB Can you please add comment what and why we need to check it here that complex?

mariusheine: @TruongQuangSB Ahh ok. Got it now. Was just confused by the namings. Can we somehow rename thinks to add the term `virtualAEA`?

PR #1859, opened 2025-09-11
Siteplan: Visualisation Cross

TruongQuangSB:


java/bundles/org.eclipse.set.model.siteplan/META-INF/MANIFEST.MF
```diff
@@ -7,7 +7,7 @@ Bundle-Version: 2.5.0.qualifier
 Bundle-ClassPath: .
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
-Bundle-RequiredExecutionEnvironment: JavaSE-21
+Bundle-RequiredExecutionEnvironment: JavaSE-17
```
mariusheine, line 10: @TruongQuangSB This change happened accidently or why have you changed it? Because in all other bundles we just JavaSE-21

mariusheine (changes requested):

PR #1883, opened 2025-09-26
Defined leading planing group

TruongQuangSB:


java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/modelservice/PlanningAccessServiceImpl.java
```diff
@@ -106,34 +117,23 @@ public Planung_Projekt getLSTPlanungProjekt(
...
-
-		// return the first ESTW group (if present)
-		if (firstESTWGroup.isPresent()) {
-			return firstESTWGroup.orElseThrow();
-		}
-
-		// return the first Geo group (if present)
-		if (firstGeoGroup.isPresent()) {
-			return firstGeoGroup.orElseThrow();
+	public Planung_Gruppe getLeadingPlanungGruppe(
```
mariusheine, line 120: @TruongQuangSB This methods throws an exception when we try to initialize a new project

image

mariusheine (changes requested):

java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/modelservice/PlanningAccessServiceImpl.java
mariusheine: @TruongQuangSB Now there seems to be another error already earlier in the new project initialization process

image

TruongQuangSB: @mariusheine i haved rebase to main

PR #1901, opened 2025-10-08
Fill blank by nil value, fix compare value by MultiColorCell

TruongQuangSB:


java/bundles/org.eclipse.set.feature.export/src/org/eclipse/set/feature/export/tablediff/CustomTableDiffService.java
```diff
@@ -135,61 +135,35 @@ private static void createDiffContent(final int i, final TableRow row,
...
+						.add(String.format(colorContent.getStringFormat(),
+								colorContent.getMultiColorValue()));
+			});
+
+			newCellContent.getValue().forEach(colorContent -> {
+				compareCellContent.getNewValue()
+						.add(String.format(colorContent.getStringFormat(),
+								colorContent.getMultiColorValue()));
+			});
+			oldCell.setContent(compareCellContent);
```
mariusheine, line 163: @TruongQuangSB This change seems pretty different from how it was before. And it seems quite unrelated to the change. What is the reason for this change?

PR #1912, opened 2025-10-13
Improve compare project by guid change

TruongQuangSB:
In this PR the `CustomDiffService` was split to two class:
 - TableStateDiffService
 - TableProjectDiffService

### DIFF
--- a/java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/diff/AbstractTableDiff.java
+++ b/java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/diff/AbstractTableDiff.java
@@ -37,7 +37,8 @@ import org.eclipse.set.services.table.TableDiffService;
 public abstract class AbstractTableDiff implements TableDiffService {
 
 	@Override
-	public Table createDiffTable(final Table firstTable, final Table secondTable) {
+	public Table createDiffTable(final Table firstTable,
+			final Table secondTable) {
 		// expand old table by new lines
 		final Table expanded = expandNewRowGroups(firstTable, secondTable);
 		TableExtensions.getTableRows(expanded)


java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/diff/AbstractTableDiff.java
```diff
@@ -0,0 +1,196 @@
...
+			if (mergedTable.getTablecontent() == null) {
+				mergedTable.setTablecontent(
+						TablemodelFactory.eINSTANCE.createTableContent());
+			}
+			mergedTable.getTablecontent().getRowgroups().add(newRowGroup);
+		}
+	}
+
+	@SuppressWarnings("static-method")
+	protected TableRow createMissingRow(final List<ColumnDescriptor> columns) {
```
mariusheine: @TruongQuangSB I would name that method `addMissingRow` or already set the correct cell values.

If you don't set the cell values here then please add a comment in `addMissingRowGroup` that we set the cell values later in `compareCell`. I would like to have this because it was very confusing in the review.

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/diff/TableProjectDiffService.java
```diff
@@ -0,0 +1,427 @@
...
+	@Override
+	protected void addMissingRowGroup(final RowGroup compareTableRowGroup,
+			final Table mergedTable) {
+		final RowGroup match = findMatchRow(compareTableRowGroup, mergedTable);
+		if (match == null) {
+			super.addMissingRowGroup(compareTableRowGroup, mergedTable);
+		}
+	}
+
+	private static RowGroup findMatchRow(final RowGroup compareTableRowGroup,
```
mariusheine: @TruongQuangSB I also would like to rename this function as it is not only a find method. It has some side effects due to `createCHangeGuidRowGroup`.

I suggest `handleOnlyChangedGUID` and return a `boolean` indicating if "normal" (superclass) handling should be performed or not.

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/diff/AbstractTableDiff.java
```diff
@@ -0,0 +1,196 @@
...
+import org.eclipse.set.ppmodel.extensions.EObjectExtensions;
+import org.eclipse.set.services.table.TableDiffService;
+
+/**
+ * Abstract class for create diff table
+ */
+public abstract class AbstractTableDiff implements TableDiffService {
+
+	@Override
+	public Table createDiffTable(final Table oldTable, final Table newTable) {
```
mariusheine: @TruongQuangSB Can it be that the parameter order must be first newTable and then oldTable? Because you are adding rows that are not existing in the old table as `PlanCompareRowType.REMOVED_ROW` in the TableProjectDiffService. In my opinion these rows should be `NEW_ROW` ...

I am bit confused 🤣

java/bundles/org.eclipse.set.model.tablemodel.extensions/src/org/eclipse/set/model/tablemodel/extensions/TableExtensions.xtend
```diff
@@ -477,16 +477,26 @@ class TableExtensions {
...
+	}
 
-			if (fc instanceof SimpleFootnoteContainer) {
-				val remarks = fc.footnotes.map[getFootnoteInfo(table, it)].
-					filterNull
+	private static def boolean isInlineFootnote(Table table,
+		FootnoteContainer fc, int maxCharInCell) {
+		if (fc instanceof SimpleFootnoteContainer) {
+			
+		}
```
mariusheine: @TruongQuangSB This can be removed then

```suggestion
```

mariusheine (changes requested):
@TruongQuangSB You did the foot note multiline fix/improvement also in this MR now, correct?

TruongQuangSB:
> @TruongQuangSB You did the foot note multiline fix/improvement also in this MR now, correct?

@mariusheine yes, the multiline problem is only by `Vergleich planung` and i haved changed here instead of `;` to `line break` als separator between `notes`

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/diff/TableProjectDiffService.java
TruongQuangSB: @mariusheine ok, this is better. But i will be still return a `RowGroup` to avoid two times `loop` to find the changed guid Row,

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/diff/AbstractTableDiff.java
TruongQuangSB: @mariusheine  i will be change to `first` and `second` 😄 . When `newTable` first, then it is no more suit to `TableState` (here should the INITAL first then the FINAL)

mariusheine (approved):

PR #1936, opened 2025-10-22
Fix Exception by Single state plan

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/overview/TableOverviewPart.java
```diff
@@ -224,6 +226,9 @@ private void calculateAllMissingTables(final IProgressMonitor monitor) {
 			tableService.transformTables(monitor, getModelSession(),
 					new HashSet<>(missingTables), TableType.FINAL,
 					controlAreaIds);
+		} else {
```
mariusheine: @TruongQuangSB But when I calculated all the missing tables it still shows "missing tables" hint on top. I would have expected to see something like everything is fine and the missing tables to be empty.

mariusheine: @TruongQuangSB I just checked in the an older snapshot. There it was behaving correctly. So you somehow broke something here 🤣

PR #1987, opened 2025-11-13
Trigger resort table after process

TruongQuangSB:


java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/sorting/CompareRouteAndKmCriterion.java
```diff
@@ -132,4 +152,34 @@ private <T> Optional<Integer> compareObj(final T first, final T second) {
...
+
+	@Override
+	public boolean isCompareDependencyOnService() {
+		return isWaitingOnService;
+	}
+
+	@Override
+	public boolean shouldTriggerComparePredicates() {
+		return geoKanteGeometryService.isFindGeometryComplete();
+	}
```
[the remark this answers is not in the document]
mariusheine: @TruongQuangSB And can't we just merge these two methods? Or in other words one predicate should be sufficient.

 ```
@Override
public boolean canPerformComparison() {
    return geoKanteGeometryService.isFindGeometryComplete();
}
```

File: java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/sorting/CompareRouteAndKmCriterion.java [hunk 2]
```diff
@@ -157,4 +162,9 @@ private <T> Optional<Integer> compareCollection(final Collection<T> first,
 		}
 		return Optional.empty();
 	}
+
+	@Override
+	public String getTriggerComparisonEventTopic() {
+		return isWaitingOnService ? Events.FIND_GEOMETRY_PROCESS_DONE : ""; //$NON-NLS-1$
```
mariusheine: @TruongQuangSB This is not working because when subscribing to the events in `ToolboxTableView#subcribeTriggerResortEvent` it initially returns an empty string and therefor never registers the event listener. 

We somehow should check it against `GeoKanteGeometryService#isFindGeometryComplete`

mariusheine: @TruongQuangSB But when is it then necessary. I checked Ssks and Sslz in PPHN and both of them are never subscribing to the event.

PR #1991, opened 2025-11-14
Optimiere Github Workflows

TruongQuangSB:


.github/workflows/ci.yml
```diff
@@ -197,12 +202,25 @@ jobs:
...
+          restore-keys: ${{ runner.os }}-m2repo-
       - name: Build
-        run: mvn -T 1.5C -U -B clean deploy checkstyle:checkstyle --settings .github/settings.xml -DaltDeploymentRepository=local::file://${{ github.workspace }}/set-m2deploy
+        run: |
+          if ([ "${{ github.event_name }}" == "pull_request" ]); then
+            # The local deployment repository is only needed for PR builds to resolve dependencies
+            mvn -T 1.5C -U -B clean deploy checkstyle:checkstyle --settings .github/settings.xml -DaltDeploymentRepository=local::file://${{ github.workspace }}/set-m2deploy
+          else
+            mvn -T 1.5C -U -B clean verify checkstyle:checkstyle --settings .github/settings.xml
+          fi
```
mariusheine: @TruongQuangSB If we start caching our maven repo now, shouldn't we then get rid of the `-U` option so that we are really using the cache?

And for what purpose did you split up for non-PR flows? I would say we still need to resolved dependencies properly also for main builds...

PR #1992, opened 2025-11-14
[PLANPRO-6021] Multiple Foundations Signal

JanErikVoigt:


java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/transform/SignalTransformator.xtend
```diff
@@ -65,9 +67,46 @@ class SignalTransformator extends BaseTransformator<Signal> {
...
+		}
+		
+		// If more then one mount like that exists, return the one with a schirm.	
+		// If there are multiple ones, take any	
+		var mount = mounts_with_schirm.head; 
+		
+		// walk up to root mount.
+		while (mount.signalBefestigung !== null) {
+			mount = mount.signalBefestigung
+		}
```
TruongQuangSB: here can be a endloss Loop, when the SignalBefestigung wrong referece, also Example: 
`Signal_Befestigung A -> Signal_Befestigung B -> Signal_Befestigung C -> Signal_Befesitung A.`

U schould check here the already found `Befesitung`

PR #1994, opened 2025-11-17
Signalinfo refactor xtend to java

JanErikVoigt:


java/bundles/org.eclipse.set.feature.siteplan/src/org/eclipse/set/feature/siteplan/transform/SignalInfo.java
```diff
@@ -0,0 +1,57 @@
...
+		final Signal_Befestigung base = this.getBaseMount();
+		if (base != null) {
+			return base.getIdentitaet().getWert();
+		}
+		return this.getFirstSignal().getIdentitaet().getWert();
+	}
+
+	// This order defines, what "First Signal" is
+	private static final Comparator<Signal> signalComparator = Comparator
+			.comparing((final Signal sig) -> sig.getIdentitaet().getWert());
```
TruongQuangSB: no need cast to `Signal` here

PR #2047, opened 2025-12-01
[PLANPRO-5832] Signal ausleger with foundation

JanErikVoigt:


java/bundles/org.eclipse.set.model.siteplan/src/org/eclipse/set/model/siteplan/impl/SiteplanPackageImpl.java
TruongQuangSB: like SignalMountType, pls make change from siteplan.ecore

PR #2051, opened 2025-12-02
feat: added control for subwork selection

linusbbb:


java/bundles/org.eclipse.set.application/src/org/eclipse/set/application/subwork/SubworkSelectionControl.java
```diff
@@ -0,0 +1,195 @@
...
+
+		final var fachdatenContainer = lst.getFachdaten();
+		if (fachdatenContainer == null) {
+			return java.util.List.of();
+		}
+
+		final var fachdaten = fachdatenContainer.getAusgabeFachdaten();
+		if (fachdaten == null) {
+			return java.util.List.of();
+		}
```
TruongQuangSB: @linusbbb all of the NPE check here can u solve with `EObjectExtensions#getNullableObject`

java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/modelservice/PlanningAccessServiceImpl.java
```diff
@@ -119,21 +112,24 @@ public Planung_Projekt getLSTPlanungProjekt(
...
 				.getLSTPlanungGruppe();
+
 		if (planingGroups.isEmpty()) {
 			return null;
 		}
 		final Map<ENUMUntergewerkArt, List<Planung_Gruppe>> groupsBySubWork = planingGroups
 				.stream()
-				.filter(group -> getUntergewerkArt(group) != null)
+				.filter(group -> getUntergewerkArt(group)
+						.equals(getCurrentUntergewerkArt()))
```
TruongQuangSB: @linusbbb here should be check one of both `Untegerwerk_Art != null`. When both of them `null` is, then should be give back the `Planung_Gruppe` in line `133`

TruongQuangSB (approved):

PR #2070, opened 2025-12-11
SWT: fix sskp dm test

mariusheine:


mariusheine:
> @TruongQuangSB You merged too fast 🤣

Hääää nevermind. IT was correct without the hyphan... Why is it failing then at me locally...

PR #2087, opened 2025-12-15
Table Overview: ignore Exception table

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceImpl.java
```diff
@@ -129,6 +131,7 @@ public final class TableServiceImpl implements TableService {
 
 	private final Map<TableCompareType, TableDiffService> diffServiceMap = new ConcurrentHashMap<>();
 	private static final Queue<Pair<BasePart, Runnable>> transformTableThreads = new LinkedList<>();
+	private static final Set<TableInfo> cantTransformationTable = new HashSet<>();
```
mariusheine: @TruongQuangSB Can we please rename that variable?

```suggestion
	private static final Set<TableInfo> nonTransformableTables = new HashSet<>();
```

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/messages/Messages.java
```diff
@@ -165,4 +165,9 @@ public class Messages {
 	 */
 	public String TableExportPart_ExportPlanningRecordsDescriptionService_ViewName;
 
+	/**
+	 * Nicht generierbar:
+	 */
+	public String TableOverviewPart_CantRendereTable;
```
mariusheine: @TruongQuangSB And rename it everywhere else 😅

mariusheine:
@TruongQuangSB When I open a non transformable table before opening the table overview I still see that table in the "Nicht überprüft" category. But I think this is not easily solvable, isn't it? Everything is fine when I then click on "überprüfen".

If it is not easily solvable then we should keep it as it is.

java/bundles/org.eclipse.set.feature.table/OSGI-INF/l10n/bundle.properties
```diff
@@ -33,4 +34,7 @@ TableExportPart_ExportInventoryDescription=Ausgabe des Zielzustands der Planung.
 TableExportPart_ExportPlanningDescription=Ausgabe des rot/gelb Vergleichs (pdf) und des Zielzustands der Planung (xlsx).
 TableExportPart_ExportStateDescription=Ausgabe des Einzelzustands der PlanPro-Datei.
 TableExportPart_FilterEmptyButton=leere Tabellen abwählen
-TableExportPart_ExportButton=Export des Dokumentensatz durchführen
\ No newline at end of file
+TableExportPart_ExportButton=Export des Dokumentensatz durchführen
+
+TableTransform_Error=Table Generierung fehler.
+TableTransform_ComparePlanError_Msg=Es gibt Fehler bei Generierung Table des Vergleichsplanung. Es wird nur die Hauptstabelle anzeigen
```
mariusheine: @TruongQuangSB 

```suggestion
TableTransform_Error=Fehler bei Tabellengenerierung
TableTransform_ComparePlanError_Msg=Es gibt Fehler bei der Tabellengenerierung der Vergleichsplanung. Es wird nur die Tabelle der Originalplanung angezeigt.
```

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceImpl.java
```diff
@@ -661,28 +668,27 @@ private IRunnableWithProgress createProgressMonitor() {
...
+						controlAreaIds);
+				while (!TableService.isTransformComplete(tableInfo, null)) {
 					Thread.sleep(2000);
-				} catch (final InterruptedException e) {
-					Thread.interrupted();
 				}
+				result.put(tableInfo, table);
+				monitor.worked(1);
+			} catch (final Exception e) {
+				Thread.interrupted();
```
mariusheine: @TruongQuangSB Shouldn't we reset the monitor in the finally clause so that it always get cleanup even in the error case?

mariusheine (changes requested):

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceImpl.java
TruongQuangSB: @mariusheine i think, it isn't nesseccary, because it will continue execute, when the thread interrupted, also the current worked is done

PR #2111, opened 2026-01-06
PlazModel: Fix PlanungsBereichValid

TruongQuangSB:


java/bundles/org.eclipse.set.feature.plazmodel/src/org/eclipse/set/feature/plazmodel/check/PlanungsBereichValid.xtend
```diff
@@ -57,32 +63,43 @@ class PlanungsBereichValid extends AbstractPlazContainerCheck implements PlazChe
...
-		val isPlanning = isPlanningObject(object)
-		val mismatchedObjects = object.referencedObjects.filter [
-			isPlanningObject(it) !== isPlanning
-		].filterNull
+		val isPlanning = isPlanningObject(source)
 
+		val mismatchedObjects = objectWithReferencesMap
+		.filter [ obj, references |
+			obj !== source && 
+			references.map[wert].exists [it == source.identitaet.wert]
```
mariusheine: @TruongQuangSB A bit easier to read in my opinion

```suggestion
			references.exists [wert == source.identitaet.wert]
```

And maybe we should combine the two filter calls (this one and the one below) or make three out of it otherwise it is a bit confusing why there are two.

mariusheine (changes requested):

PR #2126, opened 2026-01-13
[Siteplan] e2e testing

mariusheine:


web/siteplan/tests/pphn.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { expect, Page, test } from '@playwright/test'
+import pphn from './data/PPHN_1.10.0.3_01-02_Ibn-Z._-_2._AeM_2022-05-17_13-44_tg3.json' with { type: 'json' }
```
TruongQuangSB: @mariusheine when the 'json' file is static, then we can only test change in siteplan part, isn't it. It can detect the error come from java part

TruongQuangSB: but it is ok for inital state, we can use this for test of update dependency.

PR #2131, opened 2026-01-15
Fix cancel table transform issue

TruongQuangSB:


mariusheine (changes requested):

mariusheine (approved):

mariusheine:
@TruongQuangSB Can you please resolve the conflicts here. Then I can do a retest.

mariusheine:
@TruongQuangSB And here also please the conflicts... :D

mariusheine:
@TruongQuangSB When we cancel the table generation then the table is opened with the state until we have cancelled it. I would expect the table not to be opened on cancel. Also the cache should be not filled with the incomplete data.

TruongQuangSB:
> @TruongQuangSB When we cancel the table generation then the table is opened with the state until we have cancelled it. I would expect the table not to be opened on cancel. Also the cache should be not filled with the incomplete data.

fixed. pls do test again

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceImpl.java
```diff
@@ -135,8 +133,7 @@ public final class TableServiceImpl implements TableService {
 	private static final Set<TableInfo> nonTransformableTables = new HashSet<>();
 
 	private CacheService getCacheService() {
-		return ToolboxConfiguration.isDebugMode() ? Services.getNoCacheService()
-				: cacheService;
+		return cacheService;
```
mariusheine: @TruongQuangSB I think this change should be reverted. Looks like you needed it for local testing.

mariusheine (changes requested):

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceImpl.java
TruongQuangSB: @mariusheine a yes, i have forget this to revert

mariusheine (approved):

PR #2220, opened 2026-03-04
Sxxx - some adjustments

mariusheine:


java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/ToolboxTableView.java
```diff
@@ -780,6 +795,100 @@ protected ColumnGroup4HeaderLayer createGroupHeaderLayer(
...
+			return;
+		}
+		if (getMissingTables().size() == 0) {
+			final Composite parent = calculateMissingTablesPanel.getParent();
+			calculateMissingTablesPanel.dispose();
+			parent.layout(true, true);
+			parent.update();
+			calculateMissingTablesPanel = null;
+		}
+	}
```
TruongQuangSB: @mariusheine the Code here is like in TalbeOverview. I think, we can move it to TableServiceUtils ?

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceUtils.java
```diff
@@ -128,6 +136,122 @@ protected static List<TableError> getCachedTableError(
...
+		final Collection<TableInfo> missingTables = getMissingTables(
+				tableService, modelSession, controlAreaIds, tableCategory);
+		monitor.beginTask(messages.TableOverviewPart_CalculateMissingTask,
+				missingTables.size());
+		if (!modelSession.isSingleState()) {
+			// We don't need create DIFF instance for Errors detecting
+			tableService.transformTables(monitor, new HashSet<>(missingTables),
+					TableType.INITIAL, controlAreaIds);
+			tableService.transformTables(monitor, new HashSet<>(missingTables),
+					TableType.FINAL, controlAreaIds);
```
TruongQuangSB:
```suggestion
			tableService.transformTables(monitor, new HashSet<>(missingTables),
					TableType.DIFF, controlAreaIds);
```
After improve Table cache it will be always transformation table in DIFF-State, then filter Value to FINAL or INITIAL

PR #2259, opened 2026-03-13
Reactive merge function

TruongQuangSB:
- Peding Issue
  - Reload active parts after merge
  - Rendere Table in Merge Mode


mariusheine:
@TruongQuangSB There is one issue/error that we should fix before we merge this:

1. I created a complete new planing without any data inside
2. I wanted to merge it with the PPHN example project
3. I resolved all conflicts by the second planing (PPHN)
4. I clicked on save
5. I clicked on save composite planing
6. IndexOutOfBoundException occurred

TruongQuangSB:
@mariusheine it will be show a error pop-up from dialogservice when the table can't be transform and rendere a empty table, instead of the exception popup

PR #2280, opened 2026-03-20
Fix some unit test

TruongQuangSB:

PR #2281, opened 2026-03-20
Tycho pomless

TruongQuangSB:
- Require PR #2280


mariusheine:
@TruongQuangSB Can you please resolve the conflicts here. I think you are way faster with it than me ;)

TruongQuangSB:
@mariusheine done 😄

java/bundles/org.eclipse.set.core.test/testToolboxFile/.pid
mariusheine, line 1: @TruongQuangSB This file should be deleted

java/bundles/org.eclipse.set.feature.plazmodel.test/src/org/eclipse/set/feature/plazmodel/test/CalculateTopologicalCoordinateTest.java
```diff
@@ -105,7 +105,9 @@ private void thenExpectAllCoordinateAreEqualReference() {
 					.filter(entry -> entry.get(4).equals(topKanteGuid))
 					.toList();
 			assertFalse(csvEntries.isEmpty());
-			assertEquals(1, csvEntries.size());
+			assertEquals(1, csvEntries.size(), () -> {
+				return "Test";
+			});
```
mariusheine: @TruongQuangSB Just test code 😉 


```suggestion
			assertEquals(1, csvEntries.size());
```

mariusheine (changes requested):
@TruongQuangSB Why do we need to keep sine if the poms e.g. in the session service or the enumtransformation.

java/bundles/org.eclipse.set.core.test/testToolboxFile/.pid
mariusheine: @TruongQuangSB What about this file?

mariusheine:
@TruongQuangSB And why do we have to keep some of the pom files e.g. in the enum transformation, session service or pdf utils. For me there is nothing special in comparison to e.g. the pom in the feature bundle

java/bundles/org.eclipse.set.core.test/testToolboxFile/.pid
TruongQuangSB: @mariusheine deleted

TruongQuangSB:
> @TruongQuangSB Why do we need to keep sine if the poms e.g. in the session service or the enumtransformation.

was removed. i overlooked these

java/bundles/org.eclipse.set.pdf.utils/pom.xml
mariusheine, line 1: @TruongQuangSB What about this pom? It looks like this can also be deleted.

mariusheine (approved):

PR #2297, opened 2026-04-01
Implementation Sskx - Table der Tafeln

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskx/SskxTransformator.java
```diff
@@ -0,0 +1,325 @@
...
+ * Table transformation for Tabelle der Tafeln (Sskx).
+ * 
+ * @author Truong
+ */
+public class SskxTransformator extends AbstractSignalTableTransform {
+
+	private static final String SIGNAL_BEGRIFFE_BEZEICHNUNG_SEPERATOR = ", "; //$NON-NLS-1$
+	private static final List<Class<? extends Signalbegriff_ID_TypeClass>> SPEICAL_BEZEICHNUNG_HANDLE_BEGRIFFE = List
+			.of(Ne31str.class, Ne32str.class, Ne33str.class, Ne34str.class,
+					Ne35str.class);
```
mariusheine, line 67: @TruongQuangSB IS this the reason for the model change? https://github.com/eclipse-set/model/pull/57

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskx/SskxTransformator.java
```diff
@@ -0,0 +1,325 @@
...
+								+ tp.getTPBeschreibung().getWert();
+					}
+
+					if (r.getRahmenHoehe() != null
+							&& r.getRahmenHoehe().getWert() != null) {
+						return "Rahmenhöhen beachten"; //$NON-NLS-1$
+					}
+					return null;
+				}).filter(Objects::nonNull).toList(), MIXED_STRING_COMPARATOR,
+				s -> s, SIGNAL_BEGRIFFE_BEZEICHNUNG_SEPERATOR);
```
mariusheine, line 190: @TruongQuangSB How is this working? From my understanding the `fillFootnotes` will overwrite (in a later stage) the content that write into the bemerkungs column here, won't it?

Same question for the ssks, sslr, sslz and sszw transformator. IN all these transformators we call `fillFootnotes` but are also filling the Bemerkung column manually.

In the Ssvu transformator it looks like we never call `fillFootnotes`...

java/bundles/org.eclipse.set.swtbot/SWTBot.launch
mariusheine, line 1: @TruongQuangSB This looks like an accidental or debug change?

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskx/SskxTransformator.java
TruongQuangSB: @mariusheine no, the reason is the attribute `Bezeichnung` of Signal_Begriffe (DS, DV, Lange). These are fixed value and exist not in Plan file

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskx/SskxTransformator.java
TruongQuangSB: @mariusheine the `fillfootnotes` only take the notes from `Bearbeitungsvermerk` and the `fill bemerkung` are special comment for the leading object. You can see it in ppmtab.
Yeah, is ssvu should add `fillFootnotes`

java/bundles/org.eclipse.set.swtbot/SWTBot.launch
TruongQuangSB: @TruongQuangSB this ist unexpected changed 😄 .This make local execute test only the TableDataTest

mariusheine: Reverted by https://github.com/eclipse-set/set/pull/2383

PR #2320, opened 2026-04-10
Fix auto hyphenare and new line by compara value

TruongQuangSB:


java/bundles/org.eclipse.set.feature/rootdir/data/export/pdf/content.xsl
```diff
@@ -59,24 +59,26 @@
...
+						</fo:block>
+					</xsl:for-each>
 				</xsl:otherwise>
 			</xsl:choose>
 		</fo:block>
 	</xsl:template>
 
 	<xsl:template match="UnchangedValue">
-		<fo:inline hyphenate="true" language="de">
+		<fo:inline>
```
mariusheine, line 73: @TruongQuangSB This change (and also the other deletions of `hyphenate="...` were possible because the hyphenation was already enabled on the parent element? Or why do you removed them here?

mariusheine (approved):

java/bundles/org.eclipse.set.feature/rootdir/data/export/pdf/content.xsl
TruongQuangSB: @mariusheine yeah, it already defined in parent `fo:block` line 50, 63 and the `hyphenate` not work on `fo:inline`

mariusheine (approved):

PR #2327, opened 2026-04-13
Fix determin BankingInformation throught TOP_Kante Path

TruongQuangSB:


java/bundles/org.eclipse.set.application/src/org/eclipse/set/application/graph/BankServiceImpl.java
```diff
@@ -415,6 +403,53 @@ private static BankingInformation getBankingInformation(
...
+				.stream()
+				.filter(edge -> TopKanteExtensions.isConnectedTo(beginEdge,
+						edge.getValue()))
+				.findFirst()
+				.orElse(null);
+		final ID_TOP_Kante_ohne_Proxy_TypeClass connectToEnd = topPaths.stream()
+				.filter(edge -> TopKanteExtensions.isConnectedTo(endEdge,
+						edge.getValue()))
+				.findFirst()
+				.orElse(null);
```
mariusheine, line 441: @TruongQuangSB In the filter expressions we will get NullpointerExceptions if the beginEdge or endEdge are null. We should catch this somehow

java/bundles/org.eclipse.set.application/src/org/eclipse/set/application/graph/BankServiceImpl.java
mariusheine, line 1: @TruongQuangSB We should think about some unit test for the service in my opinion. I begin to loose overview how everything works with this banking lines... 😅

mariusheine (changes requested):

java/bundles/org.eclipse.set.application/src/org/eclipse/set/application/graph/BankServiceImpl.java
TruongQuangSB: @mariusheine yeah, i will do it later

mariusheine (approved):

mariusheine:
@TruongQuangSB What do you to see the plaz_model changes in #2328 ? Looks like all banking line issues are resolved now with this bugfix. Is that correct from your point of view?

TruongQuangSB:
@mariusheine yes, all was resolved by this change. The IDTopKantePath no need inclusive start and end TOP_Kanten

PR #2335, opened 2026-04-14
Ssks & Sskz Update find AEA recursive

TruongQuangSB:


java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/AussenelementansteuerungExtensions.xtend
```diff
@@ -136,4 +138,41 @@ class AussenelementansteuerungExtensions extends BasisObjektExtensions {
...
+		while (!recursiveList.nullOrEmpty) {
+			val head = recursiveList.pop
+			if (condition.test(head)) {
+				return head
+			}
+			#[head.informationPrimaer.filter(Aussenelementansteuerung),
+				head.informationSekundaer].flatten.forEach [
+				addToListFunc.apply(it)
+			]
+		}
```
mariusheine: @TruongQuangSB It would be a bit simpler to read as 


```suggestion
		val LinkedList<Aussenelementansteuerung> recursiveList = newLinkedList
		val addToListFunc = [ Aussenelementansteuerung ele |
			if (!recursiveList.exists[it === ele]) {
				recursiveList.add(ele)
			}
		]
		
		recursiveList.add(aea)
		while (!recursiveList.nullOrEmpty) {
			val head = recursiveList.pop
			if (condition.test(head)) {
				return head
			}
			#[head.informationPrimaer.filter(Aussenelementansteuerung),
				head.informationSekundaer].flatten.forEach [
				addToListFunc.apply(it)
			]
		}
```

mariusheine (changes requested):

mariusheine (approved):

PR #2352, opened 2026-04-17
Ssza: determine prefix value column N

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssza/SszaTransformator.xtend
```diff
@@ -557,4 +558,24 @@ class SszaTransformator extends AbstractPlanPro2TableModelTransformator {
 		return pointCoordinate.coordinate.
 			getStreckeKmThroughProjection(strecke).toTableDecimal
 	}
+
+	private def String getRelLagebzua(Datenpunkt dp) {
+		val lange = dp?.datenpunktAllg?.datenpunktLaenge?.wert.toTableDecimal
```
mariusheine: @TruongQuangSB 

```suggestion
		val lange = dp?.datenpunktAllg?.datenpunktLaenge?.wert?.toTableDecimal
```

Additionally there is a small typo. It should `laenge`

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssza/SszaTransformator.xtend
```diff
@@ -557,4 +558,24 @@ class SszaTransformator extends AbstractPlanPro2TableModelTransformator {
 		return pointCoordinate.coordinate.
 			getStreckeKmThroughProjection(strecke).toTableDecimal
 	}
+
+	private def String getRelLagebzua(Datenpunkt dp) {
```
mariusheine: @TruongQuangSB For better readability I would like to rename it to


```suggestion
	private def String getRelLageBZuA(Datenpunkt dp) {
```

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssza/SszaTransformator.xtend
```diff
@@ -557,4 +558,24 @@ class SszaTransformator extends AbstractPlanPro2TableModelTransformator {
...
+		if (lange === null) {
+			return ""
+		}
+
+		val direction = dp?.datenpunktAllg?.ausrichtung?.wert
+		if (direction === null) {
+			return lange
+		}
+
+		val topKante = dp.singlePoint.topKante
```
mariusheine: @TruongQuangSB If `dp` can be null we should change this to


```suggestion
		val topKante = dp?.singlePoint?.topKante
		if (topKante === null) {
			return lange
		}
```

java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/PunktObjektExtensions.xtend
```diff
@@ -257,7 +265,7 @@ class PunktObjektExtensions extends BasisObjektExtensions {
 		val result = routeThroughBereichObjekt.map [ route |
 			try {
 				return route ->
-					po.singlePoint.getStreckeKmThroughProjection(route).
+					po.singlePoints.first.getStreckeKmThroughProjection(route).
```
mariusheine, line 268: @TruongQuangSB Why do you have changed this? The `singlePoint` function delivers the same result.

java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/TopKanteExtensions.xtend
```diff
@@ -849,4 +854,49 @@ class TopKanteExtensions extends BasisObjektExtensions {
...
+				first,
+				second
+			)
+			if (compareValue > 0) {
+				downwardCount++
+			} else {
+				upwardCount++
+			}
+		}
+		return upwardCount > downwardCount
```
mariusheine, line 900: @TruongQuangSB I don't get the logic here... You are sorting the point objects by their distance. The problem is now that the kilometer on the route does not follow the same sorting or what? 

If everything is sorted consistently then updward and downward count will always be the same thus leading to `false` as return value which might not be correct...

What am I overseeing? 🤣

mariusheine (changes requested):

java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/TopKanteExtensions.xtend
TruongQuangSB: @mariusheine the Point_Objects liste was sorte by Topological distance, also `Punkt_Objekt->Punkt_Objekt_TopKante_Attribute->distance`. The Route-Kilometer are unter `Punkt_Objekt->Punkt_Objeckt_Strecke->streckenKm`. And it give situation, the the `streckenKm` isn't follow the topological (it can be wrong in the Plan). Example: the TOP_Kante A with Point:
 - A1 / top-distance 0 / route km 0
 - A2 / top-distance 10 / route km 10
 - A3 / top distnace 20 / route km 5
 - A4 / top distnace 30 / route km 30
In this case the upwardCount = 3 > downwardCount = 1 ==> The Top_Kante and the Route have same direction

 - A1 / top-distance 0 / route km 100
 - A2 / top-distance 10 / route km 90
 - A3 / top distnace 20 / route km 96
 - A4 / top distnace 30 / route km 70
In this case the upwardCount = 1 < downwardCount = 3 ==> The Top_Kante and the Route haven't same direction

mariusheine: @TruongQuangSB Ahh ok. So you are calculating what is the major trend between topological distance and strecke km. Now I got it.

java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/PunktObjektExtensions.xtend
TruongQuangSB: @mariusheine when the Punkt_Objekt more than one `Punkt_Objekt_Top_Kante`, then it will throw Exception by `singlepoint`.

mariusheine (approved):

PR #2353, opened 2026-04-20
Take final object to compare km

TruongQuangSB:


java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/sorting/CompareRouteAndKmCriterion.java
```diff
@@ -92,6 +96,32 @@ public int compare(final TableRow o1, final TableRow o2) {
...
+	 * @return the final object or the inital object, when final object not
+	 *         exist
+	 */
+	private static Ur_Objekt getCompareObjekt(final TableRow row) {
+		final Ur_Objekt obj = TableRowExtensions.getLeadingObject(row);
+		final ContainerType containerType = UrObjectExtensions
+				.getContainerType(obj);
+		if (containerType == ContainerType.FINAL) {
+			return obj;
+		}
```
mariusheine: @TruongQuangSB When displaying only the initial state we must sort by that state and not always the final state.

mariusheine (changes requested):

java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/sorting/CompareRouteAndKmCriterion.java
mariusheine: @TruongQuangSB I have implemented the table type specific sorting. Can you please have a look at my last commit?

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssit/SsitTransformationService.java
```diff
@@ -55,8 +56,9 @@ public AbstractPlanPro2TableModelTransformator createTransformator() {
 	}
 
 	@Override
-	public Comparator<RowGroup> getRowGroupComparator() {
-		return TableRowGroupComparator.builder()
+	public Comparator<RowGroup> getRowGroupComparator(
+			final TableType tableType) {
+		return TableRowGroupComparator.builder(tableType)
 				.sort("A", LEXICOGRAPHICAL, ASC) //$NON-NLS-1$
 				.build();
```
TruongQuangSB: @mariusheine no need to Override here. It is same with the `super function`

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskf/SskfTransformationService.java
```diff
@@ -62,8 +63,9 @@ public AbstractPlanPro2TableModelTransformator createTransformator() {
...
 
 	@Override
-	public Comparator<RowGroup> getRowGroupComparator() {
-		return TableRowGroupComparator.builder()
+	public Comparator<RowGroup> getRowGroupComparator(
+			final TableType tableType) {
+		return TableRowGroupComparator.builder(tableType)
 				.sort("A", MIXED_STRING, ASC) //$NON-NLS-1$
 				.build();
 	}
```
TruongQuangSB, line 71: @mariusheine same here

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskp/SskpTransformationService.java
```diff
@@ -8,7 +8,16 @@
...
+import static org.eclipse.set.feature.table.pt1.sskp.SskpColumns.Abstand_GM_2000;
+import static org.eclipse.set.feature.table.pt1.sskp.SskpColumns.Abstand_GM_2000_Bahnsteig_Anfang;
+import static org.eclipse.set.feature.table.pt1.sskp.SskpColumns.Abstand_GM_2000_Bahnsteig_Ende;
+import static org.eclipse.set.feature.table.pt1.sskp.SskpColumns.Abstand_Signal_Weiche;
+import static org.eclipse.set.feature.table.pt1.sskp.SskpColumns.Abstand_vorsignalWdh_GM_2000;
+import static org.eclipse.set.feature.table.pt1.sskp.SskpColumns.Bezugselement;
+import static org.eclipse.set.feature.table.pt1.sskp.SskpColumns.Gef_Stelle_Abstand;
+import static org.eclipse.set.feature.table.pt1.sskp.SskpColumns.H_Tafel_Abstand;
+import static org.eclipse.set.feature.table.pt1.sskp.SskpColumns.PZB_Schutzstrecke_Ist;
+import static org.eclipse.set.feature.table.pt1.sskp.SskpColumns.Wirkfrequenz;
```
TruongQuangSB: @mariusheine it's probably unexpect change

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/ssks/SsksTransformationService.java
```diff
@@ -9,7 +9,12 @@
 package org.eclipse.set.feature.table.pt1.ssks;
 
 import static org.eclipse.nebula.widgets.nattable.sort.SortDirectionEnum.ASC;
-import static org.eclipse.set.feature.table.pt1.ssks.SsksColumns.*;
+import static org.eclipse.set.feature.table.pt1.ssks.SsksColumns.Bezeichnung_Signal;
+import static org.eclipse.set.feature.table.pt1.ssks.SsksColumns.Fiktives_Signal;
+import static org.eclipse.set.feature.table.pt1.ssks.SsksColumns.Lichtraumprofil;
+import static org.eclipse.set.feature.table.pt1.ssks.SsksColumns.Reales_Signal;
+import static org.eclipse.set.feature.table.pt1.ssks.SsksColumns.Schaltkasten_Entfernung;
+import static org.eclipse.set.feature.table.pt1.ssks.SsksColumns.Ueberhoehung;
```
TruongQuangSB: @mariusheine here too

java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/sorting/CompareRouteAndKmCriterion.java
```diff
@@ -97,14 +106,15 @@ public int compare(final TableRow o1, final TableRow o2) {
...
 	 */
-	private static Ur_Objekt getCompareObjekt(final TableRow row) {
+	private static Ur_Objekt getCompareObjekt(final TableRow row,
+			final TableType tableType) {
 		final Ur_Objekt obj = TableRowExtensions.getLeadingObject(row);
-		final ContainerType containerType = UrObjectExtensions
-				.getContainerType(obj);
-		if (containerType == ContainerType.FINAL) {
+		if (tableType == null || tableType != TableType.DIFF) {
 			return obj;
```
TruongQuangSB, line 120: @mariusheine here can give  the `inital Object` back, when the table is `FINAL`, baucase by default the `leading object` is initial object. Only when the `inital object` not exsist, then it is `final`

mariusheine: @TruongQuangSB I don't get the point. For me this reads easily: Always return the leading object except for the diff table, then always return the final object if existing.

TruongQuangSB: @mariusheine yes, but when u in "Final State", it will give the "Initial object" here, when the inital & final object exists

mariusheine: @TruongQuangSB Ok but the requirement is the following:

```
TableType.INITIAL -> return the object from initial container
TableType.FINAL -> return the object from final container
TableType.DIFF -> return the object from final container, falling back to object from initial container
```

In the initial and final case I expected `obj` to already be the object from the correct container. And you say this is not the case in the final case, correct?

TruongQuangSB: @mariusheine no, the leading object is fixed after table craeted, It isn't changed by switch TableType. Only the TableCell values are change. Also
```
TableType.INITIAL -> the leading object is in inital container
TableType:FINAL -> the leading object is in initial container (here must the final object exsit)
TAbleType.DIFF -> the leading object is in initial container, when it isn't exist, then in final container
```

mariusheine: @TruongQuangSB Ahh ok. Now I got it. I will change it

mariusheine: @TruongQuangSB Please check the new implementation

TruongQuangSB: @mariusheine ok, it look fine

mariusheine (approved):

PR #2358, opened 2026-04-20
use service ranking on SetFontService

TruongQuangSB:


java/bundles/org.eclipse.set.sessionservice/src/org/eclipse/set/sessionservice/SetFontService.java
```diff
@@ -19,7 +19,7 @@
  * 
  * @author Stuecker
  */
-@Component
+@Component(property = "service.ranking:Integer=5")
```
mariusheine, line 22: @TruongQuangSB Why do you have put it? The other places where we use the "service.ranking" we added it to the ContextFunction annotations.

mariusheine (changes requested):

java/bundles/org.eclipse.set.sessionservice/src/org/eclipse/set/sessionservice/SetFontService.java
TruongQuangSB: @mariusheine on this Service haved we used both of Osgi Annotation Component and ContextFunction. But for this Service i think used Osgi Service will be better, because we used this service only for Pdf export
I haved removed the ContextFunction of this Service

mariusheine (approved):

PR #2361, opened 2026-04-23
Sskz: Add planing number

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table.pt1/OSGI-INF/l10n/bundle.properties
mariusheine: @TruongQuangSB Then we should maybe revert the accidently changed planning number for the Sxxx? See https://github.com/eclipse-set/set/pull/2319/changes

mariusheine (changes requested):

java/bundles/org.eclipse.set.feature.table.pt1/OSGI-INF/l10n/bundle.properties
TruongQuangSB: @mariusheine you are right

mariusheine (approved):

PR #2363, opened 2026-04-23
Fix table empty row by change area

TruongQuangSB:


java/bundles/org.eclipse.set.model.tablemodel.extensions/src/org/eclipse/set/model/tablemodel/extensions/CellContentExtensions.xtend
```diff
@@ -217,7 +222,9 @@ class CellContentExtensions {
 
 	static def dispatch Iterable<String> getStringValueIterable(
 		StringCellContent content) {
-		return content.value
+		return content?.value?.filterNull.map[trim].filter [
```
mariusheine: @TruongQuangSB Then we should consistently do

```suggestion
		return content?.value?.filterNull?.map[trim]?.filter [
```

java/bundles/org.eclipse.set.model.tablemodel.extensions/src/org/eclipse/set/model/tablemodel/extensions/CellContentExtensions.xtend
```diff
@@ -188,7 +188,12 @@ class CellContentExtensions {
 
 	static def dispatch String getPlainStringValue(
 		CompareStateCellContent content) {
-		return '''«content.oldValue.stringValueIterable»/«content.newValue.stringValueIterable»'''
+		val oldValue = content.oldValue.stringValueIterable
+		val newValue = content.newValue.stringValueIterable
```
mariusheine: @TruongQuangSB And here as well

```suggestion
		val oldValue = content?.oldValue?.stringValueIterable
		val newValue = content?.newValue?.stringValueIterable
```

mariusheine (changes requested):

mariusheine (approved):

PR #2364, opened 2026-04-23
Fix  cached table with hourglass cells

TruongQuangSB:


java/bundles/org.eclipse.set.utils.table/src/org/eclipse/set/utils/table/TableModelInstanceBodyDataProvider.java
TruongQuangSB, line 1: The function in this class was moved to TableServiceUtils

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/internal/TableServiceImpl.java
```diff
@@ -483,9 +497,19 @@ private void saveTableToCache(final Table table,
...
+					return;
 				}
+				tableChangedData.computeIfPresent(tableInfo, (key, value) -> {
+					if (value.isEmpty()) {
+						return Collections.emptyList();
+					}
+					fillDelayCells(TableExtensions.getTableRows(table), value,
+							TableType.DIFF);
+					return Collections.emptyList();
+				});
```
TruongQuangSB, line 512: Update hourglass cells before storage table to cache

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/ToolboxTableView.java
```diff
@@ -749,15 +754,14 @@ void export() {
...
-						&& t.getName().startsWith(tableInfo.shortcut())
+						&& t.getName().startsWith(extractShortcut())
 						&& t.isAlive())
 				.toList();
 		if (!transformatorThreads.isEmpty() && !getDialogService()
 				.confirmExportNotCompleteTable(getToolboxShell())) {
 			return;
 		}
-		final Map<TableType, Table> tables = compileService.compile(tableInfo,
-				getModelSession(), controlAreaIds);
```
TruongQuangSB, line 760: use current rendere table to export instead of compile again

TruongQuangSB: currently the Execel-Export export only inital or final table, therefore we need the compile here.

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/ToolboxTableView.java
```diff
@@ -247,6 +247,11 @@ private void outdatedUpdate() {
 		}
 	}
 
+	private String extractShortcut() {
+		final String elementId = getToolboxPart().getElementId();
+		return elementId.substring(elementId.lastIndexOf(".") + 1); //$NON-NLS-1$
+	}
```
mariusheine, line 252: @TruongQuangSB We should reuse the existing method `extractShortcut` from the TableService.

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/ToolboxTableView.java
```diff
@@ -323,7 +329,7 @@ protected void preDestroy() {
 		ToolboxEvents.unsubscribe(getBroker(), tableDataChangeHandler);
 		ToolboxEvents.unsubscribe(getBroker(), selectionControlAreaHandler);
 		getBroker().unsubscribe(secondaryPlanningLoadedHanlder);
-		getBroker().send(Events.CLOSE_PART, tableInfo.shortcut());
+		getBroker().send(Events.CLOSE_PART, extractShortcut());
```
mariusheine, line 331: @TruongQuangSB Why not using the tableInfo here anymore?

java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/events/TableDataChangeEvent.java
mariusheine: @TruongQuangSB From above changes I think, that we expect the tableShortcut here instead of the full name is that correct? Maybe we should rename the field as well.

mariusheine (changes requested):

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/ToolboxTableView.java
TruongQuangSB: @mariusheine because the `tableInfo` can be `null`

java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/events/TableDataChangeEvent.java
TruongQuangSB: @TruongQuangSB yeah, this is only the tabel shortcut

mariusheine (approved):

PR #2380, opened 2026-04-28
Fix NPE by find route km

TruongQuangSB:


java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/PunktObjektExtensions.xtend
```diff
@@ -242,10 +242,10 @@ class PunktObjektExtensions extends BasisObjektExtensions {
 				poGuid) as List<Pair<String, List<String>>>
 			return cachedValue.flatMap[value].toList
 		}
-
-		if (po.punktObjektStrecke.size == 1 &&
-			routeThroughBereichObjekt.contains(
-				po.punktObjektStrecke.first.IDStrecke.value)) {
+		val poRoute = po?.punktObjektStrecke?.first?.IDStrecke?.value
+		if (po.punktObjektStrecke.size == 1 && poRoute !== null &&
```
mariusheine, line 249: @TruongQuangSB Then we can simplify this to

```suggestion
		if (poRoute !== null &&
```

java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/PunktObjektExtensions.xtend
```diff
@@ -265,7 +265,7 @@ class PunktObjektExtensions extends BasisObjektExtensions {
 		val result = routeThroughBereichObjekt.map [ route |
 			try {
 				return route ->
-					po.singlePoints.first.getStreckeKmThroughProjection(route).
+					po?.singlePoints?.first?.getStreckeKmThroughProjection(route).
```
mariusheine: @TruongQuangSB And here should also add 

```suggestion
					po?.singlePoints?.first?.getStreckeKmThroughProjection(route)?.
```

java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/PunktObjektTopKanteExtensions.xtend
```diff
@@ -215,7 +215,7 @@ class PunktObjektTopKanteExtensions extends BasisObjektExtensions {
 	def static List<Strecke> getStreckenThroughBereichObjekt(
 		Punkt_Objekt_TOP_Kante_AttributeGroup potk) {
 		val topPoint = new TopPoint(potk)
-		return potk.container.strecke.filter [ route |
+		return potk?.container.strecke.filter [ route |
```
mariusheine: @TruongQuangSB Here we should do


```suggestion
		if (potk?.container?.strecke === null) {
			return null
		}
		return potk.container.strecke.filter [ route |
```

mariusheine (changes requested):

java/bundles/org.eclipse.set.ppmodel.extensions/src/org/eclipse/set/ppmodel/extensions/PunktObjektExtensions.xtend
TruongQuangSB: @mariusheine y must still check, it give only one `Punkt_Objekt_Strecke`. The `poRoute` take only first element, but the list can be more than one element

mariusheine: @TruongQuangSB Ok but then we need to do at least


```suggestion
		if (po?.punktObjektStrecke?.size == 1 && poRoute !== null &&
```

TruongQuangSB: @mariusheine i have added null check by start of function. Therefore no need to check `po` null here anymore. And the `po.punkObjektStrecke` can't be, when the `po != null`

mariusheine: @TruongQuangSB Perfect. This looks nice now!

mariusheine (approved):

PR #2417, opened 2026-05-06
Correct Table Excel template

TruongQuangSB:


java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/export/xsl/AbstractTransformTableHeader.java
```diff
@@ -163,12 +163,13 @@ protected Element transformTable() {
...
-		for (int i = 0, columNumber = 1; i <= getHeaderLastColumnIndex(
-				sheet); i++, columNumber++) {
+		int columNumber = 1;
+		for (int i = 0; i <= getHeaderLastColumnIndex(sheet); i++) {
 			final Pair<Integer, Float> pair = transformColumn(cols, sumWidth, i,
 					columNumber);
 			columNumber = pair.getKey().intValue();
 			sumWidth = pair.getValue().floatValue();
+			columNumber++;
 		}
```
mariusheine, line 173: @TruongQuangSB This change does not change anything in the result. Or do I oversee something?

java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/export/xsl/AbstractTransformTableHeader.java
```diff
@@ -234,6 +237,7 @@ protected Set<Element> transformRowCells(final XSSFRow row) {
 				i += transformSpanCell(cell, excelCell);
 				addCell(cells, rowNum, cell, i);
 			}
+			i++;
 		}
```
mariusheine, line 241: @TruongQuangSB Same here. It is logically the same as before

mariusheine (changes requested):

java/bundles/org.eclipse.set.utils/src/org/eclipse/set/utils/export/xsl/AbstractTransformTableHeader.java
TruongQuangSB: @mariusheine  i have restructur for "Code Recommend of SonarQueb", because `columnNumber` can be changed inside the loop

mariusheine (approved):

PR #2441, opened 2026-05-18
Update Vue components to composition API

linusbbb:


TruongQuangSB:
@linusbbb do you have forgot another *.vue file to modify or this is wrong branch 😄 . I'm sure, that give more than 10 *.vue files

TruongQuangSB (changes requested):
see comment above

mariusheine:
Follow up PR is #2464

PR #2450, opened 2026-05-19
Ssks: fix wrong general condition

TruongQuangSB:


mariusheine:
Became obsolete by new requirements which are implemented in #2478

PR #2526, opened 2026-07-07
No export of changes in excel

mariusheine:


java/bundles/org.eclipse.set.model.tablemodel.extensions/src/org/eclipse/set/model/tablemodel/extensions/CellContentExtensions.xtend
```diff
@@ -213,16 +208,36 @@ class CellContentExtensions {
...
+	static def dispatch boolean isEmpty(Void content) {
+		return true
+	}
+
+	static def dispatch boolean isEmpty(StringCellContent content) {
+		return content.value.empty
+	}
+
+	static def dispatch boolean isEmpty(CompareStateCellContent content) {
+		return content?.newValue?.stringValueIterable.isNullOrEmpty
```
TruongQuangSB, line 232:
```suggestion
		return content?.newValue?.isEmpty
```
@mariusheine the `newValue` is also `CellContent` type, therefore you can directly test with the `isEmpty` funciton

java/bundles/org.eclipse.set.model.tablemodel.extensions/src/org/eclipse/set/model/tablemodel/extensions/CellContentExtensions.xtend
```diff
@@ -213,16 +208,36 @@ class CellContentExtensions {
...
+	}
+
+	static def dispatch boolean isEmpty(MultiColorCellContent content) {
+		return content.value.empty
+	}
+
+	static def dispatch boolean isEmpty(CompareTableCellContent content) {
+		return content.mainPlanCellContent?.plainStringValue.isNullOrEmpty
```
TruongQuangSB, line 240:
```suggestion
		return content.mainPlanCellContent?.isEmpty
```
@mariusheine  same here

java/bundles/org.eclipse.set.model.tablemodel.extensions/src/org/eclipse/set/model/tablemodel/extensions/CellContentExtensions.xtend
```diff
@@ -197,12 +197,7 @@ class CellContentExtensions {
 
 	static def dispatch String getPlainStringValue(
 		CompareStateCellContent content) {
-		val oldValue = content?.oldValue?.stringValueIterable
-		val newValue = content?.newValue?.stringValueIterable
-		if (oldValue.isNullOrEmpty && newValue.nullOrEmpty) {
-			return ""
-		}
-		return '''«oldValue»/«newValue»'''
+		return content?.newValue?.plainStringValue;
```
TruongQuangSB, line 200: @mariusheine i think, this function should give both value back. We used also for PDF export

TruongQuangSB (changes requested):

java/bundles/org.eclipse.set.model.tablemodel.extensions/src/org/eclipse/set/model/tablemodel/extensions/CellContentExtensions.xtend
mariusheine: @TruongQuangSB No it is not used for pdf export. For PDF the compare content is handled properly by FOP.

java/bundles/org.eclipse.set.model.tablemodel.extensions/src/org/eclipse/set/model/tablemodel/extensions/CellContentExtensions.xtend
mariusheine: @TruongQuangSB Ok but then I need to change the dispatch for StringCellContent to 

```
static def dispatch boolean isEmpty(StringCellContent content) {
	return content.stringValueIterable.isNullOrEmpty
}
```

so that emptiness is properly checked. Or what do you think?

TruongQuangSB: @mariusheine the isn't neccessary, the `content.value` already list of String. Therefore we can't direct check `nullOrEmpty`. However we can filter the `blank value` first

java/bundles/org.eclipse.set.model.tablemodel.extensions/src/org/eclipse/set/model/tablemodel/extensions/CellContentExtensions.xtend
TruongQuangSB: @mariusheine yeah you are right. But i think it better to give both of value back. It it easy to compare, if two CompareCell have same value. We used it also in `TableDataProvider` to compare cell.
For the Excel export we can simple take only the `newValue`

TruongQuangSB:
in https://github.com/eclipse-set/set/pull/2531 resolve

PR #2553, opened 2026-07-15
PDF-Export: add info text in page with compare plan cell

TruongQuangSB:


java/bundles/org.eclipse.set.feature/rootdir/data/export/pdf/common.xsl
```diff
@@ -208,7 +208,27 @@ http://www.eclipse.org/legal/epl-v20.html
...
+									<fo:table-cell>
+										<fo:block/>
+									</fo:table-cell>
+								</fo:table-row>
+								<fo:table-row height="10mm">
+									<fo:table-cell display-align="after">
+										<fo:block xsl:use-attribute-sets="significant-information-style">
+											<fo:retrieve-marker retrieve-class-name="ComparePageText"
+																					retrieve-boundary="page"
+																					retrieve-position="first-starting-within-page"/>
```
mariusheine, line 226: @TruongQuangSB If you do an export e.g.the Ssks where you this multi page table thing, then the marker is detected across all sub pages. This results in the hint showing on page a although there is only a change on page b. Can you properly detect this as well?

mariusheine: @TruongQuangSB Ahh and if the footnotes are on a seperate page and have compare changes then the marker is also missing

mariusheine (changes requested):

java/bundles/org.eclipse.set.feature/rootdir/data/export/pdf/common.xsl
TruongQuangSB: @mariusheine the each page of Ssks rendere complete table, because the maxium width of the A3-Page will only page "A" or  "B" visualtion. Therefore the hint will be showed at both page. I think, for the first should we leave it like this, the renovation transformation  in another Issue, whenn needed

mariusheine (approved):

PR #2557, opened 2026-07-16
Always open validation part by single state plan

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/export/ExportPlanningRecordsDescriptionService.java
```diff
@@ -57,7 +56,7 @@ public PartDescription getDescription(final IEclipseContext context) {
 				// toolboxViewType
 				getToolboxViewGroup(),
 				// defaultForNature
-				PlanProFileNature.INFORMATION_STATE,
+				null,
```
mariusheine, line 59: @TruongQuangSB What is the reason for this change? It looks unrelated to me...

mariusheine (changes requested):

java/bundles/org.eclipse.set.feature.table/src/org/eclipse/set/feature/table/export/ExportPlanningRecordsDescriptionService.java
TruongQuangSB: @mariusheine here was defined, this part should be default for the single state plan, when the plan valid ist

mariusheine: @TruongQuangSB Ahh now I got it. And what is happening when there is no default found for a nature? To which view are we falling back?

mariusheine: @TruongQuangSB Ok I answered it myself 🤣 We simply show nothing then.

mariusheine (approved):

PR #2570, opened 2026-07-21
Signal Tabelle: update determin side distance

TruongQuangSB:


java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskz/SskzTransformator.java
```diff
@@ -253,19 +252,13 @@ private void fillTrackMitteDistance(final TableRow row,
...
+					if (sideDistance.getDistanceToMainTrack() != null) {
+						if (sideDistance.getDistanceToNeighborTrack() != null) {
 							addTopologicalCell(row, getColumn(cols,
 									SskzColumns.Abstand_FEAx_Gleismitte));
 						}
-						return String.format("%s%s", //$NON-NLS-1$
-								Math.abs(sideDistance.getFirst()),
-								trackDistance.isEmpty() ? "" //$NON-NLS-1$
-										: " " + trackDistance); //$NON-NLS-1$
+						return sideDistance.toString();
```
mariusheine, line 261: @TruongQuangSB This would also output the "x" in the Sskz. If this is desired/expected we should document this also in the ppmtab files

mariusheine (changes requested):

java/bundles/org.eclipse.set.feature.table.pt1/src/org/eclipse/set/feature/table/pt1/sskz/SskzTransformator.java
TruongQuangSB: @mariusheine the `toString()` function of SignalSideDistance will output the `x`, when given't distance

mariusheine: @TruongQuangSB Ahh ok, and because you check it above it will never be printed... Got it

mariusheine (approved):

PR #2574, opened 2026-07-23
Table overview export

TruongQuangSB:


java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/dialogservice/SelectionMultiValueDialog.java
```diff
@@ -0,0 +1,187 @@
...
+	private void updateButtonOnSelection() {
+		final long selectedCount = Arrays.stream(listView.getTable().getItems())
+				.filter(TableItem::getChecked)
+				.count();
+		final int totalCount = listView.getTable().getItemCount();
+		getButton(IDialogConstants.SELECT_ALL_ID)
+				.setEnabled(selectedCount < totalCount);
+		getButton(IDialogConstants.DESELECT_ALL_ID)
+				.setEnabled(selectedCount > 0);
+		getButton(IDialogConstants.OK_ID).setEnabled(selectedCount > 0);
```
mariusheine: @TruongQuangSB The dialog buttons are a bit confusing. In my opinion we should get rid of the cancel button because we can not really cancel the export. When we deselected everything there might still be tables to be exported. This implies that we just have the OK button and we never disable it.

TruongQuangSB: @mariusheine hmm, you are right, for this case we don't need the `cancel` button. I will make this configurationable

java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/dialogservice/SelectionMultiValueDialog.java
mariusheine: Here we must use now `...setEnabled(cancelButtonLabel == null || selectedCount > 0);`

mariusheine (changes requested):

java/bundles/org.eclipse.set.core/src/org/eclipse/set/core/dialogservice/SelectionMultiValueDialog.java
TruongQuangSB: ok

mariusheine (approved):

PR #2597, opened 2026-08-10
TableStyle test: fix not all table was tested

TruongQuangSB:


java/pom.xml
```diff
@@ -315,6 +315,12 @@
 									<ws>win32</ws>
 									<arch>x86_64</arch>
 								</environment>
+								<!-- Only for osgi plugin test -->
+								<environment>
+									<os>linux</os>
+									<ws>gtk</ws>
+									<arch>x86_64</arch>
+								</environment>
```
TruongQuangSB, line 323: this need for the osgi plugin test, when we build the set in Linux

mariusheine (approved):

